PR #20, opened 2015-01-11
Fix error using when eq/ord with mutual-rec type

runoshun:
- Add tests for mutual recursive type to  src_test/test_deriving_eq.ml,
   src_test/test_deriving_ord.ml
 - Fix code generation for list type in eq and ord plugin to avoid the
   error of 'let rec' restriction (This kind of expression is not
   allowed as right-hand side of `let rec')


whitequark:
Thanks!

PR #21, opened 2015-01-12
Handle type extensions

zoggy:


src/ppx_deriving.mli
```diff
@@ -18,15 +18,34 @@ open Parsetree
...
+
+(** Creating {!deriver} structure. *)
+val create :
+  string ->
+  ?core_type: (core_type -> expression) ->
+  ?type_ext_str: (options:(string * expression) list -> path:string list ->
+                   type_extension -> structure) ->
+  ?type_ext_sig: (options:(string * expression) list -> path:string list ->
+                   type_extension -> signature) ->
+  type_decl_str: (options:(string * expression) list -> path:string list ->
```
whitequark: Actually, why not make these optional too, for uniformity?

src/ppx_deriving.mli
```diff
@@ -155,6 +177,14 @@ val fold_right_type_decl : (string -> 'a -> 'a) -> type_declaration -> 'a -> 'a
 val fold_type_decl : ('a -> string -> 'a) -> 'a -> type_declaration -> 'a
 [@@ocaml.deprecated "Use fold_left_type_decl"]
```
whitequark: Since we're breaking the external interface, this can be removed.

src/ppx_deriving.mli
```diff
@@ -121,13 +140,16 @@ val path_of_type_decl : path:string list -> type_declaration -> string list
 (** [mangle_type_decl ~fixpoint affix type_] derives a function name from [type_] name
     by doing nothing if [type_] is named [fixpoint] (["t"] by default), or
     appending or prepending [affix] via an underscore. *)
```
whitequark: appending and/or prepending

src_plugins/ppx_deriving_create.ml
```diff
@@ -114,10 +114,12 @@ let sig_of_type ~options ~path ({ ptype_loc = loc } as type_decl) =
...
 
 let () =
-  Ppx_deriving.(register deriver {
-    core_type = None;
-    structure = (fun ~options ~path type_decls ->
-      [Str.value Nonrecursive (List.concat (List.map (str_of_type ~options ~path) type_decls))]);
-    signature = (fun ~options ~path type_decls ->
-      List.concat (List.map (sig_of_type ~options ~path) type_decls));
-  })
+  Ppx_deriving.(register
```
whitequark: The indentation in `Ppx_deriving.register` blocks is inconsistent in the standard derivers. Can you reformat?

whitequark:
After the changes I mentioned, should be ready to merge.

src/ppx_deriving.mli
zoggy: I wonder what's the use of the rewriter if it does not handle type declarations. That's no problem for me, but making handling of type declarations mandatory makes it impossible to forget such a parameter, which can sometimes be a stupid error hard to think about.

src_plugins/ppx_deriving_create.ml
zoggy: Ok. Which one do you prefer ?

whitequark:
```
Ppx_deriving.(register (create deriver
  ~type_decl_str:(fun ...)
  ~type_decl_sig:(...)
  ()
))
```

src/ppx_deriving.mli
whitequark: I can imagine derivers that only want to derive core types, or derivers that never work in signatures.

My point is, there's no conceptual difference between showing an error in `type_ext_str` and `type_decl_str`, why make one of these awkward?

zoggy: Ok. Then, the default behaviour for these would be to create no items, rather than fail as the default type_ext_sig and type_ext_decl ?

whitequark: No, the default behavior would of course be the same, to fail.

zoggy:
Is it ok, now ?

PR #26, opened 2015-03-26
Allow derivation of ord and show for mutually recursive types with polymorphic variable

Twinside:
Related to whitequark/ppx_deriving#23

It works by adding a type signature to every generated function for both of these derivers, with strong polymorphic constraint.


whitequark:
Do you think you could add a test? Maybe with more than one variable as well.

Twinside:
Sure, but later tonight

Twinside:
I've added eq in the mix, to be more consistent.

whitequark:
Thanks!

PR #27, opened 2015-03-27
Simplifying generated code for ord

Twinside:
Currently the generated code for ord looks like that:
~~~~{ocaml}
type v =
  | Foo
  | Bar of int* string
  | Baz of string
  [@@deriving ord]

let rec compare_v : v -> v -> int=
  fun lhs  ->
    fun rhs  ->
      match (lhs, rhs) with
      | (Foo ,Foo ) -> 0
      | (Bar (lhs0,lhs1),Bar (rhs0,rhs1)) ->
          (match (fun (a : int)  -> fun b  -> Pervasives.compare a b) lhs0
                   rhs0
           with
           | (-1)|1 as x -> x
           | _ ->
               (match (fun (a : string)  -> fun b  -> Pervasives.compare a b)
                        lhs1 rhs1
                with
                | (-1)|1 as x -> x
                | _ -> 0))
      | (Baz lhs0,Baz rhs0) ->
          (match (fun (a : string)  -> fun b  -> Pervasives.compare a b) lhs0
                   rhs0
           with
           | (-1)|1 as x -> x
           | _ -> 0)
      | _ ->
          let to_int = function | Foo  -> 0 | Bar (_,_) -> 1 | Baz _ -> 2 in
          ((fun (a : int)  -> fun b  -> Pervasives.compare a b)) (to_int lhs)
            (to_int rhs)
~~~~

With this patch, it looks like the following.

~~~~{ocaml}
let rec compare_v : v -> v -> int=
  fun lhs  ->
    fun rhs  ->
      match (lhs, rhs) with
      | (Foo ,Foo ) -> 0
      | (Bar (lhs0,lhs1),Bar (rhs0,rhs1)) ->
          (match Pervasives.compare lhs0 rhs0 with
           | 0 -> Pervasives.compare lhs1 rhs1
           | x -> x)
      | (Baz lhs0,Baz rhs0) -> Pervasives.compare lhs0 rhs0
      | _ ->
          let to_int = function | Foo  -> 0 | Bar (_,_) -> 1 | Baz _ -> 2 in
          Pervasives.compare (to_int lhs) (to_int rhs)
~~~~

The first big difference is the removal of an unneeded last match
case, coming from a fold_left use, and replaced by a `reduce_compare`
function taking the first expression as base case.

Thus, this snippets:

~~~~{ocaml}
(match (fun (a : string)  -> fun b  -> Pervasives.compare a b) lhs0
         rhs0
 with
 | (-1)|1 as x -> x
 | _ -> 0)
~~~~

Becomes

~~~~{ocaml}
(fun (a : string)  -> fun b  -> Pervasives.compare a b) lhs0 rhs0
~~~~

The second optimisation is to eta-reduce all the call to
Pervasives.compare, as their types are well known by the compiler,
they don't bring anything to the table. The previous example become

~~~~{ocaml}
Pervasives.compare lhs0 rhs0
~~~~

Finally, there is a change in the comparison operator:

~~~~{ocaml}
match %e with
| (-1|1) as x -> x
| _ -> %rest
~~~~

is transformed to

~~~~{ocaml}
match %e with
| 0 -> %rest
| x -> x
~~~~

This is due to the fact that the Pervasives.compare function doesn't
guarantee you to obtain `(-1|1)` as output value if you're greater or
lower, but a negative or positive number. Only the equality case is
specified, being 0, so it's safer to match against it when we're calling
non-derived comparison functions. This change affect the documentation
in the README stating that the function return -1, 0 or 1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ The function `fprintf` is locally defined in the printer.
 Plugins: eq and ord
 -------------------
 
-_eq_ derives a function comparing values by semantic equality; structural or physical depending on context. _ord_ derives a function defining a total order for values, returning `-1`, `0` or `1`. They're similar to `Pervasives.(=)` and `Pervasives.compare`, but are faster, allow to customize the comparison rules, and never raise at runtime. _eq_ and _ord_ are short-circuiting.
+_eq_ derives a function comparing values by semantic equality; structural or physical depending on context. _ord_ derives a function defining a total order for values, returning a negative value if lower, `0` if equal or a positive value is greater. They're similar to `Pervasives.(=)` and `Pervasives.compare`, but are faster, allow to customize the comparison rules, and never raise at runtime. _eq_ and _ord_ are short-circuiting.
 
 ``` ocaml
 # type t = [ `A | `B of int ] [@@deriving eq, ord];;
--- a/src_plugins/ppx_deriving_ord.ml
+++ b/src_plugins/ppx_deriving_ord.ml
@@ -20,12 +20,16 @@ let argn kind =
   Printf.sprintf (match kind with `lhs -> "lhs%d" | `rhs -> "rhs%d")
 
 let compare_reduce acc expr =
-  [%expr match [%e expr] with (-1|1) as x -> x | _ -> [%e acc]]
+  [%expr match [%e expr] with 0 -> [%e acc] | x -> x]
+
+let reduce_compare = function
+  | [] -> [%expr 0]
+  | x :: xs -> List.fold_left compare_reduce x xs
 
 let wildcard_case int_cases =
   Exp.case [%pat? _] [%expr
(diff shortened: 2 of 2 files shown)


whitequark:
Are you sure the eta-reduction you perform is correct? I've deliberately added a lambda with a type annotation. I think I verified this with an assembly listing, though that was a while ago and I don't remember.

The other change is OK.

Twinside:
The cases where the Pervasives.compare is used is when we are matching against a known "atomic" type or when comparing constructor using to_int. In both case the type is monomorphic with the arguments, so it lets all the latitude to the compiler to call an optimized version.

And the function `expr_of_type` still fulfill the contract, it returns a function being able to compare two elements.

whitequark:
Indeed you're right; I've verified with -dcmm on 4.02.1.

PR #29, opened 2015-04-01
Adding handling of shadowed standard types in mutually recursive types.

Twinside:
Shadowing standard type, while definitely bad style, will result in invalid generated code using ppx_deriving (and I know a code base with those...):

~~~~{ocaml}
type es =
  | ESBool of bool
  | ESString of string
and bool =
  | Bfoo of int * ((int -> int) [@compare fun _ _ -> 0])
and string =
  | Sfoo of String.t * ((int -> int) [@compare fun _ _ -> 0])
  [@@deriving ord{ allow_std_type_masking }]
~~~~

will call `Pervasives.compare` for `ESBool` and `ESString`, yielding to a crash at runtime. The first solution implemented in this patch is to reject derivation of shadowed types. If you known what you're doing, you can use the `allow_std_type_masking` to force derivation (if possible), and try to call the good functions while generating code for the mutually recursive types (like the one added in the examples).

While testing, I found that some warning were generated with types of only one constructor, due to the unreached wildcard pattern, there is a fix for this.

All of this implemented for ord/eq/show

### DIFF
--- a/src/ppx_deriving.ml
+++ b/src/ppx_deriving.ml
@@ -5,6 +5,8 @@ open Parsetree
 open Ast_helper
 open Ast_convenience
 
+module StringSet = Set.Make(String)
+
 type deriver = {
   name : string ;
   core_type : (core_type -> expression) option;
@@ -268,4 +270,37 @@ let binop_reduce x a b =
 let strong_type_of_type ty =
   let free_vars = free_vars_in_core_type ty in
   Typ.force_poly @@ Typ.poly free_vars ty
+
+let predefined_types =
+  [
+    "unit";
+    "int";
+    "int32";
+    "int64";
+    "nativeint";
+    "float";
+    "bool";
+    "char";
+    "string";
+    "bytes"
+  ]
+
+let predefined_set =
+  List.fold_right StringSet.add predefined_types StringSet.empty
+
+let extract_typename_of_type_group deriver ~allow_shadowing type_list =
+  let add_name acc ty =
+    let typename = ty.ptype_name.txt in
+    let is_shadowing_predefined =
+      StringSet.mem typename predefined_set in
+    if is_shadowing_predefined && not allow_shadowing then
+      raise_errorf
+        ~loc:ty.ptype_loc
+        ("%s don't allow derivation of shadowed standard type %s. " ^^
+         "Use option 'allow_std_type_masking' to lift the restriction.")
+        deriver
+        typename
+    else
+      StringSet.add ty.ptype_name.txt acc in
+  List.fold_left add_name StringSet.empty type_list
 
--- a/src/ppx_deriving.mli
+++ b/src/ppx_deriving.mli
@@ -2,6 +2,8 @@
 
 open Parsetree
 
+module StringSet : Set.S with type elt = string
+
 (** {2 Registration} *)
 
 (** A type of deriving plugins.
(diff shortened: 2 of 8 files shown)


whitequark:
Shouldn't it be called `allow_std_type_shadowing`?

The code looks good, but I don't quite understand how it works. Could you post an example of broken and corrected generated code?

Twinside:
Ah yes, I will rename the option, for the broken code:

~~~~{ocaml}
type es =
  | ESBool of bool
  | ESString of string
and bool =
  | Bfoo of int* ((int -> int)[@compare fun _  -> fun _  -> 0])
and string =
  | Sfoo of String.t* ((int -> int)[@compare fun _  -> fun _  -> 0])
[@@deriving ord]
~~~~
Was generating:
~~~~{ocaml}
let rec compare_es : es -> es -> int=
  fun lhs  ->
    fun rhs  ->
      match (lhs, rhs) with
                                   (* this is wrong, we want compare_bool *)
      | (ESBool lhs0,ESBool rhs0) -> Pervasives.compare lhs0 rhs0
                                   (* this is wrong, we want compare_string *)
      | (ESString lhs0,ESString rhs0) -> Pervasives.compare lhs0 rhs0
      | _ ->
          let to_int = function | ESBool _ -> 0 | ESString _ -> 1 in
          Pervasives.compare (to_int lhs) (to_int rhs)
and compare_bool : bool -> bool -> int=
  fun lhs  ->
    fun rhs  ->
      match (lhs, rhs) with
      | (Bfoo (lhs0,lhs1),Bfoo (rhs0,rhs1)) ->
          (match Pervasives.compare lhs0 rhs0 with
           | 0 -> ((fun _  -> fun _  -> 0)) lhs1 rhs1
           | x -> x)
      | _ ->
          let to_int = function | Bfoo (_,_) -> 0 in
          Pervasives.compare (to_int lhs) (to_int rhs)
    (* while not wrong, this pattern is useless because we have only one
        constructor, and it generate a warning during compilation *)
and compare_string : string -> string -> int=
  fun lhs  ->
    fun rhs  ->
      match (lhs, rhs) with
      | (Sfoo (lhs0,lhs1),Sfoo (rhs0,rhs1)) ->
          (match String.compare lhs0 rhs0 with
           | 0 -> ((fun _  -> fun _  -> 0)) lhs1 rhs1
           | x -> x)
      | _ ->
          let to_int = function | Sfoo (_,_) -> 0 in
          Pervasives.compare (to_int lhs) (to_int rhs)
    (* while not wrong, this pattern is useless because we have only one
        constructor, and it generate a warning during compilation *)
~~~~

And now generates (when using the option):

~~~~{ocaml}
let rec compare_es : es -> es -> int=
  fun lhs  ->
    fun rhs  ->
      match (lhs, rhs) with
      | (ESBool lhs0,ESBool rhs0) -> compare_bool lhs0 rhs0
      | (ESString lhs0,ESString rhs0) -> compare_string lhs0 rhs0
      | _ ->
          let to_int = function | ESBool _ -> 0 | ESString _ -> 1 in
          Pervasives.compare (to_int lhs) (to_int rhs)
and compare_bool : bool -> bool -> int=
  fun lhs  ->
    fun rhs  ->
      match (lhs, rhs) with
      | (Bfoo (lhs0,lhs1),Bfoo (rhs0,rhs1)) ->
          (match Pervasives.compare lhs0 rhs0 with
           | 0 -> ((fun _  -> fun _  -> 0)) lhs1 rhs1
           | x -> x)
and compare_string : string -> string -> int=
  fun lhs  ->
    fun rhs  ->
      match (lhs, rhs) with
      | (Sfoo (lhs0,lhs1),Sfoo (rhs0,rhs1)) ->
          (match Pervasives.compare lhs0 rhs0 with
           | 0 -> ((fun _  -> fun _  -> 0)) lhs1 rhs1
           | x -> x)
~~~~

This work in two phases, first we build a set of the name of all the types declared in a mutually recursive way, if a standard type is detected, an error is raised. If the option is present, we try to generate, and when we find a type name present in the set of type name, we force the generation for user types, instead of the standard one.

whitequark:
Ok, got it. It's a rather niche case but I have no objections since the code seems clear enough.

Twinside:
The option renaming is now included in the pull request

whitequark:
Note: I've reverted this and implemented proper hygiene instead. The behavior is still accessible if you use e.g. `bool [@nobuiltin]` explicitly.

PR #32, opened 2015-04-09
Adding installation without Opam

Twinside:
This pull request update topkg with the latest one, which include a patch to detect the '.lib' extension generated under windows with Ocaml + msvc.

It also add an install command to the makefile to be able to install the ppx without Opam on windows and linux, related to whitequark/ppx_deriving#16

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,6 @@ install:
 		awk '{ print $$1 }' | \
 		sed -e 's:"?*::g'  | \
 		xargs ocamlfind install ppx_deriving
-	cp `ocamlfind query ppx_deriving`/ppx_deriving_main.native `ocamlfind query ppx_deriving`/ppx_deriving$(EXE)
+	mv `ocamlfind query ppx_deriving`/ppx_deriving_main.native `ocamlfind query ppx_deriving`/ppx_deriving$(EXE)
 
 .PHONY: gh-pages release
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+include $(shell ocamlc -where)/Makefile.config
+
 build:
 	cp pkg/META.in pkg/META
 	ocaml pkg/build.ml native=true native-dynlink=true
@@ -45,5 +47,13 @@ release:
 	opam publish prepare $(NAME_VERSION) $(ARCHIVE)
 	opam publish submit $(NAME_VERSION)
 	rm -rf $(NAME_VERSION)
+
+install:
+	ocamlfind remove ppx_deriving
+	grep -E '^[[:space:]]+' ppx_deriving.install | \
+		awk '{ print $$1 }' | \
+		sed -e 's:"?*::g'  | \
+		xargs ocamlfind install ppx_deriving
+	cp `ocamlfind query ppx_deriving`/ppx_deriving_main.native `ocamlfind query ppx_deriving`/ppx_deriving$(EXE)
 
 .PHONY: gh-pages release
--- a/pkg/topkg.ml
+++ b/pkg/topkg.ml
@@ -27,8 +27,14 @@ module type Exts = sig
   val interface_opt : string list
   (** [interface_opt] is [".cmx" :: interface] *) 
 
+  val c_library : string list
+  (** Extension of the native platform library, [".a"] for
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -46,4 +48,12 @@ release:
 	opam publish submit $(NAME_VERSION)
 	rm -rf $(NAME_VERSION)
 
+install:
+	ocamlfind remove ppx_deriving
+	grep -E '^[[:space:]]+' ppx_deriving.install | \
+		awk '{ print $$1 }' | \
+		sed -e 's:"?*::g'  | \
+		xargs ocamlfind install ppx_deriving
+	cp `ocamlfind query ppx_deriving`/ppx_deriving_main.native `ocamlfind query ppx_deriving`/ppx_deriving$(EXE)
```
whitequark: mv?

Makefile
```diff
@@ -46,4 +48,12 @@ release:
 	opam publish submit $(NAME_VERSION)
 	rm -rf $(NAME_VERSION)
 
+install:
+	ocamlfind remove ppx_deriving
+	grep -E '^[[:space:]]+' ppx_deriving.install | \
```
whitequark: This is... a rather heavy-handed way.

Twinside: The goal was to avoid information duplication on the file to copy, do you have another idea in mind?

Makefile
Twinside: no problem

Makefile
whitequark: Any chance you could implement a tool using `opamlib`'s .install file parser? This was what I was planning some time ago, it would also get rid of cygwin dependency.

Twinside: I've used this Makefile against MSYS (not cygwin), and as ocamlbuild needs a bash in the path to work well (it was a harsh discovery), I don't think we can completely get rid of all Unix replacement platform on windows :-/

vbmithr: On 10/04/2015 09:40, Vincent wrote:
> I've used this Makefile against MSYS (not cygwin), and as ocamlbuild
> needs a bash in the path to work well (it was a harsh discovery), I
> don't think we can completely get rid of all Unix replacement platform
> on windows :-/

Maybe try using ocp-build. It's a work in progress, and I would not
recommand it for everyday use, but it has been designed to work on Windows.

-- 
Vincent

Twinside: I've looked after ocp-build, I have big hopes for it, but right now there is a clear lack of documentation for the tool which makes it a no-go for an use in a professional context.

vbmithr: On 10/04/2015 12:22, Vincent wrote:
> I've looked after ocp-build, I have big hopes for it,

Abandon all hope, ye who enter here…

but right now
> there is a clear lack of documentation for the tool which makes it a
> no-go for an use in a professional context.

Yes, obviously. That's why I said "try to", just in case it would work
here, you would be done with it :)

-- 
Vincent

whitequark: ocp-build is garbage and it will be nowhere near my code. ocamlbuild does not need bash as far as I'm aware (or even if it does, it should be considered a bug); though /some/ ocamlbuild scripts/plugins explicitly could.

There is however a good argument that we're in a makefile, so we need some kind of make. I neither want to ensure that my makefiles can be read with nmake nor add a separate batch or nmake file, so merging.

Twinside:
Fixing moving the executable instead of copying.

And Ocamlbuild use bash on windows as seen in this [comment](https://github.com/ocaml/ocaml/blob/cf6714ead24163382ab73426a8206421e0d7c706/ocamlbuild/shell.mli#L18) and on this [win32 shell call](https://github.com/ocaml/ocaml/blob/cbfe627f925ab2bab93bae7a7bc9f6ee6afb8637/ocamlbuild/my_std.ml#L260), **sigh**

And it is an already known bug, from this [mantis ticket](http://caml.inria.fr/mantis/print_bug_page.php?bug_id=4981), the related bugs are interesting. If I motivate myself, I may look after patching ocamlbuild someday.

whitequark:
Gross. ocamlbuild already implements an elaborate command syntax that should obviate the need to invoke shell entirely.

Issue #33, opened 2015-04-19
Recursive application of the mapper


paurkedal:
I think there should be a way to apply the deriving-mapper to expression-valued attributes. Maybe as an alternative Arg.expr or by exposing the mapper as a global?

My motivation is that I wrote a [syntax extension to generate random instances of types](https://github.com/paurkedal/ppx_deriving_random).  The derived-from type takes an attribute `[@random ...]` to override with custom generator code, where I need reliable way to refer to generators for parameters or the type itself, e.g. in
```ocaml
type 'a term =
  | Var of 'a
  | App  [@weight 0.5] of 'a term * 'a term
  | Lam of ('a * 'a term) [@random random_lambda [%random: 'a term]]
  [@@deriving random]
```
In other cases `random_lambda` might use just `[%random: 'a]`.  The internal name `poly_a` is needed in either case, which makes me think the best way to refer to it is to use `[%random: ...]`.  I could systematically apply all parameter generators to the custom generator and use:
```ocaml
  | Lam of ('a * 'a term) [@random random_lambda random_term]
```
In my opinion, explicit arguments are nicer.

whitequark:
Nice syntax extension!

I agree that the mapper should be exported. Would you make a PR?

paurkedal:
Thanks!

Sure, though I can't think of a nice way to do it without breaking backwards compatibility.  If doing it from scratch I would have replaced the ~option ~path parameters with a single private record for extensibility.  To keep backwards compatibility exposing it as a global seems the best option.  I think it needs to be a reference set by the mapper function, may disguised as a function:
```ocaml
val get_mapper : unit -> mapper
(** [get_mapper ()] is the current deriving-mapper.  This must be called from a deriving plugin. *)
```
Any other suggestion?

whitequark:
I think a (more) elegant solution would be moving Ppx_deriving_main.{derive,derive_type_decl,derive_type_ext,module_from_input_name,mapper} into Ppx_deriving. mapper would be just a field, since it's almost stateless*. Ppx_deriving_main would only consist of code that does plugin loading and maintaining the Ast_mapper cookie.

(*) It has implicit state in form of module_nesting. It could be tempting to make that state local to the mapper, but in fact it /should/ be global, since the whole compiler-libs are global, and I can't really envision using the mapper outside of a compiler-libs context. It has to rely on Location.input_name anyway.

paurkedal:
[...]

PR #34, opened 2015-04-19
Expose the mapper to derivers, moving some code out of main.

paurkedal:
Ad issue #33, maybe some final adjustments?

### DIFF
--- a/src/ppx_deriving.ml
+++ b/src/ppx_deriving.ml
@@ -304,3 +304,157 @@ let extract_typename_of_type_group deriver ~allow_shadowing type_list =
       StringSet.add ty.ptype_name.txt acc in
   List.fold_left add_name StringSet.empty type_list
 
+let derive path pstr_loc item attributes fn arg =
+  let deriving = find_attr "deriving" attributes in
+  let deriver_exprs, loc =
+    match deriving with
+    | Some (PStr [{ pstr_desc = Pstr_eval (
+                    { pexp_desc = Pexp_tuple exprs }, []); pstr_loc }]) ->
+      exprs, pstr_loc
+    | Some (PStr [{ pstr_desc = Pstr_eval (
+                    { pexp_desc = (Pexp_ident _ | Pexp_apply _) } as expr, []); pstr_loc }]) ->
+      [expr], pstr_loc
+    | _ -> raise_errorf ~loc:pstr_loc "Unrecognized [@@deriving] annotation syntax"
+  in
+  List.fold_left (fun items deriver_expr ->
+      let name, options =
+        match deriver_expr with
+        | { pexp_desc = Pexp_ident name } ->
+          name, []
+        | { pexp_desc = Pexp_apply ({ pexp_desc = Pexp_ident name }, ["",
+            { pexp_desc = Pexp_record (options, None) }]) } ->
+          name, options |> List.map (fun ({ txt }, expr) ->
+            String.concat "." (Longident.flatten txt), expr)
+        | { pexp_loc } ->
+          raise_errorf ~loc:pexp_loc "Unrecognized [@@deriving] option syntax"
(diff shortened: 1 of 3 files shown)


src/ppx_deriving.ml
```diff
@@ -2,6 +2,7 @@ open Longident
 open Location
 open Asttypes
 open Parsetree
+open Ast_mapper
```
whitequark: You don't have to open the Ast_mapper, all references are qualified

On the other hand you can even remove Ast_helper below

src/ppx_deriving.ml
```diff
@@ -304,3 +305,181 @@ let extract_typename_of_type_group deriver ~allow_shadowing type_list =
...
+  | Some _ -> assert false
+  | None -> []
+
+let add_plugins plugins =
+  let loaded  = get_plugins () in
+  let plugins = List.filter (fun file -> not (List.mem file loaded)) plugins in
+  List.iter dynlink plugins;
+  let loaded  = loaded @ plugins in
+  Ast_mapper.set_cookie "ppx_deriving"
+    (Exp.tuple (List.map (fun file -> Exp.constant (Const_string (file, None))) loaded))
```
whitequark: All Dynlink-related stuff should live in Ppx_deriving_main. Ppx_deriving and the library shouldn't even depend on Dynlink

src/ppx_deriving.ml
paurkedal: They are both used multiple places in `ppx_deriving`, esp. I would prefer to keep Ast_mapper due to the record field uses.  OTOH, I could remove all open statements from `ppx_deriving_main`.

src/ppx_deriving.ml
paurkedal: Are you sure?  There is a match on @@@findlib.ppxopt in the `structure` local function which calls `add_plugins` and which calls `dynlink`.  I guess it's there to support the `#ppx_deriving` directives in the toplevel.

src/ppx_deriving.ml
whitequark: Oh. Maybe just add Ast_mapper.{} around that? I would really prefer not to have get_cookie and such in global scope, it's dirty enough anyway.

src/ppx_deriving.ml
whitequark: Oh, yeah, that should also go into `_main`. `Ppx_deriving` itself should really be free of any ocaml/ocamlc-related invocation issues, that would make reusing it much cleaner.

whitequark: Basically, `Ppx_deriving` is an AST mapper, `Ppx_deriving_main` is an AST mapper *specifically for the default OCaml toolchain*.

paurkedal:
I've qualified all Ast_mapper reference and moved dynlink stuff back to main.  In doing the latter, I changed recursion of structure and signature so that they use the mapper.  That allowed me to hook in the ppxopt from main.  I'm not sure how to test the toplevel support though.  I guessed `#ppx "deriving"` but I'm not sure how ppxopt is passed.

whitequark:
Nope. Toplevel is tested in the exact same way as non-toplevel: with findlib. You do `#require "ppx_deriving.std";` for example.

paurkedal:
That explains why `#ppx` is not documented. Digging around a bit, I see that `topfind.ml.in` injects `@@@findlib.ppxopt` to communicate with the syntax extension.

whitequark:
I verified it works on toplevel. Merging, thanks!

whitequark:
It is an overlook that `#ppx` is not documented. It is equivalent to a documented `ocamlc -ppx` option.

paurkedal:
Right, I did test that, as well. And I also have uncomitted code which uses the mapper.

Thanks!

Issue #39, opened 2015-06-17
regression with 4.02.2 regarding attributes


c-cube:
Looks like attributes are ignored in 4.02.2:

```ocaml
#require "ppx_deriving.show";;
type foo = { bar : int [@opaque] } ;;
show_foo {bar=1};;
```

expect: `{ bar = <opaque>"}`
got: `{ bar = 1 }`

whitequark:
Yeah, the code above now looks like `type foo = { bar [@opaque] : int }`. Can you please make a PR?

c-cube:
I will, tomorrow. I assume you mean that the attribute is attached to the field label, rather than the field type.

whitequark:
Exactly. I have asked for this change, in fact.

c-cube:
I'm not exactly sure how to proceed. Apparently, the previous design (attaching attributes to types) allowed to use `expr_of_typ` on the type. Now, in the case of a record field, I would have to either duplicate attribute detection in `expr_of_type` for records, or give the field attributes to `expr_of_typ`…

vbmithr:
```
utop # type foo = { bar [@opaque] : int } [@@deriving show];;                                                                      
Error: Syntax error
```

It's me, or ?

whitequark:
Not you. In 4.02.2, the attribute can't be placed after record field name to avoid ambiguity. The attribute after the type goes into its slot instead. (Think of it as moving the attribute from `{ bar * : int }` to `{ bar : int * }`.

whitequark:
@c-cube Can you detect the attribute in expr_of_type for records and attach it back to the record type so that dependent code does not have to change or distinguish these cases?

PR #40, opened 2015-06-19
merge attributes of record fields and types (for OCaml 4.02.2)

c-cube:
First attempt at fixing #39. Strategy: for every record field `{label : ty}`, put attributes of `label` into `ty`.

### DIFF
--- a/src_plugins/ppx_deriving_create.ml
+++ b/src_plugins/ppx_deriving_create.ml
@@ -20,8 +20,8 @@ let attr_split attrs =
   Ppx_deriving.(attrs |> attr ~deriver "split" |> Arg.get_flag ~deriver)
 
 let find_main labels =
-  List.fold_left (fun (main, labels) ({ pld_type; pld_loc } as label) ->
-    if Ppx_deriving.(pld_type.ptyp_attributes |>
+  List.fold_left (fun (main, labels) ({ pld_type; pld_loc; pld_attributes } as label) ->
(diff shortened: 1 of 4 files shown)


whitequark:
LGTM. The tests will pass but that's irrelevant since Travis does not run them on 4.02.2.

By the way, would you consider becoming a co-maintainer of ppx_deriving?

c-cube:
I tested on 4.02.2 (I'm not totally sure every attribute is tested though). Becoming maintainer: why not, I understand you don't want to spend too much time on OCaml now.

c-cube:
looks like I got myself a few new repositories :)

whitequark:
Yeah, I've been a bit surprised by just how many

c-cube:
Ok then, shall I merge this PR myself? Do you want me to release a bugfix version?

whitequark:
Sure; I trust your judgement so you can work on these projects as you see fit.

A bugfix release is definitely in order. The older ones should probably be marked as broken on opam, too.

c-cube:
I see 2.1 in the changelog but not in tags, is that normal? I plan to only release 2.0.1 as a bugfix release.

whitequark:
2.1 is the next release, not released yet. I don't believe in patch versions in semver so I release in a simple major.minor fashion: major is incremented when there is API breakage (not limited to types), minor is incremented when there is not.

c-cube:
Let's 2.1 then.

PR #45, opened 2015-07-20
new plugin: ppx_deriving_lens

didier-wenzek:
This proposed plugin adds a lens to each field of a record
To be used with https://github.com/avsm/ocaml-lens/

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,25 @@ val create_record :
   record
 ```
 
+Plugin: lens
+------------
+
+_lens_ is a plugin that generates lenses for all fields of a record.
+
+``` ocaml
+type car = {
+  make : string;
+  model: string;
+  mileage: int;
+} [@@deriving lens];;
+
+val car_make: (car, string) Lens.t
+val car_model: (car, string) Lens.t
+val car_mileage: (car, int) Lens.t
+```
+
+These lenses can than be used with the opam `lens` package.
+
 Building ppx drivers
 --------------------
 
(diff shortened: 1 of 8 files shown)


Drup:
Why put it in ppx_deriving ? It should go in the lens package.

didier-wenzek:
I think that the implementation is by far more related to ppx_deriving than to lens.
I agree that the plugin can be used if no Lens package is defined with the proper signature.
An alternative is to put the code in a specific package as for ppx_deriving_yojson.
Is it what you mean ?

c-cube:
I agree that it probably should go in `lens`. There might be other lens types — e.g. using (possibly first-class) modules — that would need a different ppx to work.

whitequark:
While I think the plugin is great, the policy of the `ppx_deriving` package is that it will only depend on stdlib (I consider stdlib to include `ppx_tools` and `ocamlfind`). Thus `deriving(lens)` cannot be added. I agree with other commenters that it should preferably go in the `lens` package, or in `ppx_deriving_lens` if that is not possible.

rgrinberg:
@didier-wenzek How come you aren't using the more maintained ocaml-lens fork? https://github.com/pdonadeo/ocaml-lens

didier-wenzek:
@rgrinberg You are right the opam package references the fork of pdonadeo.

didier-wenzek:
Done, the plugin has been merged into https://github.com/pdonadeo/ocaml-lens

Drup:
Happy ending ! :3

PR #55, opened 2015-09-24
Eta expand equality functions

choeger:
This is necessary in case of recursive groups, due to
the let-rec restriction to syntactic functions.

Such a change is possibly necessary for other plugins
as well.

Signed-off-by: Christoph Höger <christoph.hoeger@tu-berlin.de>


whitequark:
Looks good--do you think you could do the same change for Ord?

choeger:
Not right now, but it shouldn't be a big issue. I'll have a look when I find some time for it.

whitequark:
I also did this for ord and show, so don't bother.

PR #56, opened 2015-09-24
Link ppx_deriving_main with linkall

diml:
The ppx\_deriving binary should be linked with `-linkall` so that plugins are free to use modules of the stdlib that are not referenced by ppx_deriving itself.

### DIFF
--- a/_tags
+++ b/_tags
@@ -3,6 +3,7 @@ true: warn(@5@8@10@11@12@14@23@24@26@29@40), bin_annot, safe_string, debug
 
 "src": include
 <src/*.{ml,mli,byte,native}>: package(dynlink), package(compiler-libs.common), package(ppx_tools.metaquot)
+<src/ppx_deriving_main.{byte,native}>: linkall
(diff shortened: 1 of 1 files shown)


diml:
Could you do a release of ppx\_deriving with this patch?

Without it ppx\_deriving fails to load all of our ppx deriving plugins (ocaml/opam-repository#4884) as we are using `StdLabels` which is not used by ppx\_deriving.

whitequark:
@diml Will do in a moment.

I've reviewed your ppx source drop and I think your work is of exceptional quality; can I mention that in the ppx_deriving 3.0 annotation email or should I wait yet?

whitequark:
Done: https://github.com/ocaml/opam-repository/pull/4885. I've restarted the build...

diml:
Sure, feel free to mention that. Note that it's not all from me though, several people from Jane Street worked on this code.

whitequark:
Yeah, I meant "your" as in Jane Street.

PR #60, opened 2015-11-03
Fix META with ocp-build

dinosaure:
`ocp-build` does not handle the `+=` in META file.

### DIFF
--- a/pkg/META.in
+++ b/pkg/META.in
@@ -25,9 +25,7 @@ package "main" (
 package "std" (
   version = "%{version}%"
   description = "Meta-package for all built-in derivers"
-  requires  = "ppx_deriving.show ppx_deriving.eq ppx_deriving.ord"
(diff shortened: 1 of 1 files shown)


whitequark:
I'm pretty sure this needs to be fixed in ocp-build.

dinosaure:
Yes, I think :) . I try that today. If you want to close this PRs, it's OK.

whitequark:
Let's wait until it's clear what the ocp-build outcome is.

whitequark:
@dinosaure Was there any progress with ocp-build?

dinosaure:
I fixed locally ocp-build, and OCamlPro merged my PR but I don't know when is the next release :( ! So, you can close this PR, I can use my PR instead.

whitequark:
OK, thanks!

Issue #65, opened 2015-12-21
Why does ppx_deriving.create always has a `unit` at the end?


rgrinberg:
for example:

```
type point = { x: float ; y: float } [@@deriving create]

val create : x:float -> y:float -> unit -> point
```

I thought the `unit` is only necessary if any of the labels have a default value? (i.e. optional).

whitequark:
True, but I didn't implement that for reason that I am unable to remember. That would be a breaking change... worth it?

rgrinberg:
I'm against breaking people's code if there's a bunch of usage in the wild. No idea whether that's the case however.

Another possibility is to have `[@@deriving make]` that will not generate the extra `unit` when it's not necessary. Might address #62 as well? This would be satisfactory to me.

whitequark:
Sure. Would you prepare a PR?

rgrinberg:
Sure, I'll give it a try. First, let me try fixing `create` outright to see if there was anything concrete that stopped you.

whitequark:
I've already looked, there's nothing in particular, it just doesn't have the check for optionals. Should be easy enough to fix

rgrinberg:
@whitequark I think we can close this. Can we expect a release in the near future?

whitequark:
3.1 released opam publish submit ./ppx_deriving.3.1

PR #66, opened 2015-12-22
Very crude attempt at fixing create #65

rgrinberg:
So I managed to hack something together that seems to work but I need some
guidance on how to proceed. Currently I've changed `create` to asses the scope
of the changes but now I'm not sure how to proceed with a real `make` plugin.
Seems like almost all of the code will be the same between the 2 plugins, how
should I share the code between the two?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 _deriving_ is a library simplifying type-driven code generation on OCaml >=4.02.
 
-_deriving_ includes a set of useful plugins: [show][], [eq][], [ord][eq], [enum][], [iter][], [map][iter], [fold][iter], [create][], [yojson][], [protobuf][].
+_deriving_ includes a set of useful plugins: [show][], [eq][], [ord][eq], [enum][], [iter][], [map][iter], [fold][iter], [make][], [yojson][], [protobuf][].
 
 Sponsored by [Evil Martians](http://evilmartians.com).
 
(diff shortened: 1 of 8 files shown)


whitequark:
LGTM. Just copy the code instead (with modifications), since we're not really interested in improving the old version it in any way, and please add a deprecation warning to the emitted code of `[@deriving create]`.

rgrinberg:
@whitequark I've created ppx_deriving.make but I'm not so sure about the deprecation warning for the emitted code. It's likely that sers of ppx_deriving.create are exposing those functions to users of their own code and just deprecating seems excessive all of a sudden. Are we sure about this?

whitequark:
Hmmmm true. Just replace it in README then.

arvidj:
Sorry for digging this up. I just tried to figure out form which opam library the `create` deriver comes from. I was pretty confused since it wasn't mentioned in the README.md of `ppx_deriving`. Then I was very confused when I saw that there were both `make` and `create`, with no discussion of the difference. I think mentioning `create` as a deprecated deriver in the README.md would be a good idea.

gasche:
Done: https://github.com/ocaml-ppx/ppx_deriving/commit/077b26d11947dbdfb8f6aca3944c8c3d97ed37ac.

arvidj:
> Done: [077b26d](https://github.com/ocaml-ppx/ppx_deriving/commit/077b26d11947dbdfb8f6aca3944c8c3d97ed37ac).

Thanks!

Issue #68, opened 2016-01-27
`ppx_deriving.make` not found by ocamlfind


labichn:
The package descriptor for the new `make` plug-in is missing (should be in pkg/META.in). This causes the following error when ppx_deriving.std is referenced:

> ocamlfind: Package `ppx_deriving.make' not found - required by `ppx_deriving.std'

Pull request incoming.

PR #69, opened 2016-01-27
Missing pkg/META.in description for `ppx_deriving.make`.

labichn:
See issue #68.


whitequark:
Thanks!

By the way generally there is no need to open a separate issue if you already have a PR incoming, since PRs can have everything issues have... just to save you the work next time.

PR #78, opened 2016-02-24
Honor the [@printer] annotation for variant types.

johanneskloos:
This patch makes ppx_deriving_show honor [@printer] annotations
on the cases of a variant type, closing
https://github.com/whitequark/ppx_deriving/issues/77.

In particular, consider the following code:

type t =
  | First [@printer fun fmt _ -> Format.fprintf fmt "first case"]
  | Second
      [@@deriving show]

let () =
  Format.printf "%a, %a@." pp First pp Second

It now outputs "first case, Printer_variant.Second".


whitequark:
Looks good but please make indentation consistent with the rest of that file.

whitequark:
Thanks!

PR #82, opened 2016-03-10
Fix type problems in 'map' and 'iter'.

andrewray:
The signature generated for `map` was less general than the actual
implementation ie

```
type 'a t = ...
val map : ('a -> 'a) -> 'a t -> 'a t
```

as opposed to

```
val map : ('a -> 'b) -> 'a t -> 'b t
```

A bit of new machinary is added to ppx_deriving.ml to support this
`core_type_with_fresh_vars`.

For `iter` the signature was actually more general than the
implementation

```
type ('a,'b) t = ...
val iter : ('a -> unit) -> ('b -> unit) -> ('a,'b) t -> unit
```

whereas the implementation gave

```
val iter : ('a -> unit) -> ('a -> unit) -> ('a,'a) t -> unit
```

This was due to a type constraint on the `Ptyp_var` that forced
all the poly arg functions to the same type.

### DIFF
--- a/src/ppx_deriving.cppo.ml
+++ b/src/ppx_deriving.cppo.ml
@@ -313,6 +313,23 @@ let core_type_of_type_decl { ptype_name = { txt = name }; ptype_params } =
 let core_type_of_type_ext { ptyext_path ; ptyext_params } =
   Typ.constr ptyext_path (List.map fst ptyext_params)
 
+let core_type_with_fresh_vars bound type_decl = 
+  let vars,bound = 
+    List.fold_right
+      (fun _ (vars,bound) -> 
+        let v = fresh_var bound in
+        (Typ.var v::vars,v::bound)) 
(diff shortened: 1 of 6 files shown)


whitequark:
Tests?

rgrinberg:
I didn't look too closely but shouldn't this be a problem with `fold` as well?

andrewray:
`fold` seems to be ok.

```
utop # module X : sig type ('a,'b) t = { a : 'a; b : 'b }[@@deriving fold] end = 
                  struct type ('a,'b) t = { a : 'a; b : 'b }[@@deriving fold] end;;
module X : sig type ('a, 'b) t = { a : 'a; b : 'b; } val fold : ('c -> 'a -> 'c) -> ('c -> 'b -> 'c) -> 'c -> ('a, 'b) t -> 'c end     
``

PR #85, opened 2016-03-15
Result support

rgrinberg:
For early feedback


rgrinberg:
@whitequark do you mind checking out the last commit. It's yielding this strange error:

```
File "src_test/test_deriving_show.cppo.ml", line 199, characters 0-74:
Error: This expression has type (string/2286, int) Result.result
       but an expression was expected of type
         (string/3, int) Ppx_deriving_runtime.Result.result =
           (string/3, int) Result.result
       Type string/2286 is not compatible with type string/3
Command exited with code 2.
```

I'm not quite sure what's the problem here.

whitequark:
The error message indicates different types with the same name. It would seem that the `string` at `test_ppx_deriving_show.cppo.ml:199` differs from the one in `Pervasives`. Your guess is as good as mine as to why... sorry, no time right now to debug it.

rgrinberg:
@whitequark never mind, it was just a silly bug with the `string` type being shadowed in the `show` test.

The PR seems ready now.

whitequark:
Why did you add `result` support in `eq` but not any other plugins?

rgrinberg:
I added it in show as well. I can add it to other plugins however. Any other ones you think are important?

whitequark:
Show, sure, that's obvious. But why eq? Are you expecting `result` to be embedded in other types? If that is generally useful, then it should be added to all plugins, just like the other standard type. If that isn't very useful, then I'm wondering why bother with eq...

rgrinberg:
To be honest, I just added it to the plugins I use most often. I'm open to adding it to other plugins as well, I'm just not as familiar with them. I guess it makes sense for `ord`, `iter`, `map`, and `fold` as well. Should I add them too?

whitequark:
Yes, and please update the doc as well.

rgrinberg:
@whitequark I've added more plugins.

whitequark:
Yes, in README.

whitequark:
Also, I think you missed `ord`?

rgrinberg:
@whitequark ord + readme note done.

whitequark:
re: README, I meant to extend the "Plugin conventions" section... just add it to the list.

rgrinberg:
@whitequark done

whitequark:
Wonderful, thank you!

PR #86, opened 2016-03-19
Change Arg api to use Result.result

### LINKED ISSUE #75: Change Arg to use Result
There's no need for the

```ocaml
[`Ok of 'a | `Error of string]
```

polymorphic variant as `result` is available in 4.03 and as a findlib packages for earlier OCaml's.

Is there anyway to make this transition without breaking a bunch of code?

rgrinberg:
Previously it would use a polymorphic variant version of result. This is
no longer necessary as OCaml now has a standard result type.

The sanitize change will come in a different PR

Fix #75

### DIFF
--- a/src/ppx_deriving.mli
+++ b/src/ppx_deriving.mli
@@ -72,35 +72,42 @@ val string_of_core_type : Parsetree.core_type -> string
     The [~name] argument is used in error messages and should receive
     the name of the deriving plugin, e.g. ["show"]. *)
 module Arg : sig
+  (** A type of conversion functions.
+
+      A conversion function of type ['a conv] converts a raw expression into an
+      argument of type ['a]. Or returns [Result.Error "error"] if conversion
+      fails. *)
+  type 'a conv = expression -> ('a, string) Result.result
+
   (** [expr] returns the input expression as-is. *)
-  val expr : expression -> [> `Ok of expression ]
+  val expr : expression conv
 
   (** [bool expr] extracts a boolean constant from [expr], or returns
-      [`Error "boolean"] if [expr] does not contain a boolean literal. *)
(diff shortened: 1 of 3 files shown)


rgrinberg:
Does the META file need to be updated? I've already included result as a dependency for ppx_deriving_runtime in the META in a previous PR but perhaps there are other places where we need it?

src/ppx_deriving.mli
```diff
@@ -72,35 +72,39 @@ val string_of_core_type : Parsetree.core_type -> string
     The [~name] argument is used in error messages and should receive
     the name of the deriving plugin, e.g. ["show"]. *)
 module Arg : sig
+  (** Convert raw expression into an argument. Returns [Result.Error "error"]
+      if conversion fails *)
```
whitequark: `A type of conversion functions.` etc, please see above how it's written for the type `deriver`.

whitequark:
LGTM other than doc nit.

whitequark:
> Does the META file need to be updated?

Ah yes. ppx_deriving.api should also require the result package.

rgrinberg:
Fixed META and doc.

src/ppx_deriving.mli
```diff
@@ -72,35 +72,42 @@ val string_of_core_type : Parsetree.core_type -> string
     The [~name] argument is used in error messages and should receive
     the name of the deriving plugin, e.g. ["show"]. *)
 module Arg : sig
+  (** A type of conversion functions.
+
+      A conversion function of type ['a conv] converts a raw expression into an
+      argument of type ['a]. Returns [Result.Error "error"] if conversion fails
+  *)
```
whitequark: Much better but there's a dot missing at the end, also `or returns` sounds nicer.

whitequark:
Thanks!

PR #92, opened 2016-04-06
minor: per-constructor, per-field location in equality code

gasche:
I'm currently trying to use ppx_deriving in a codebase that uses some
GADTs, and running into the problem that the generated code does not
type-check. One extra difficulty is that the locations in the
generated code are wrong. For example, consider

    type _ expr =
    | Int : int -> int expr
    | Prod : 'a expr * 'b expr -> ('a * 'b) expr
    [@@deriving eq]

the current implementation will report

    type _ expr =
    ^^^^^^^^^^^^^
    File "foo.ml", line 1, characters 0-98:
    Error: This pattern matches values of type (ex#0 * ex#1) expr
           but a pattern was expected which matches values of type int expr
           Type ex#0 * ex#1 is not compatible with type int

after the present commit, one gets instead

    | Prod : 'a expr * 'b expr -> ('a * 'b) expr
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "foo.ml", line 3, characters 0-44:
    Error: This pattern matches values of type (ex#0 * ex#1) expr
           but a pattern was expected which matches values of type int expr
           Type ex#0 * ex#1 is not compatible with type int

which allows to pinpoint the problematic constructor, and thus orient
ourselves in the -dsource output to further understand the typing error.

Note that the handling of location is extremely heavy-handed and
rather inelegant (global side-effect with rollback). The fault for
this seems to lie in ppx-tools that does not follow Ast_helper's
convention of having ?loc parameters.

### DIFF
--- a/src_plugins/ppx_deriving_eq.cppo.ml
+++ b/src_plugins/ppx_deriving_eq.cppo.ml
@@ -148,7 +148,8 @@ let str_of_type ~options ~path ({ ptype_loc = loc } as type_decl) =
     | Ptype_abstract, Some manifest -> expr_of_typ quoter manifest
     | Ptype_variant constrs, _ ->
       let cases =
-        (constrs |> List.map (fun { pcd_name = { txt = name }; pcd_args } ->
+        (constrs |> List.map (fun { pcd_name = { txt = name }; pcd_args; pcd_loc } ->
+          with_default_loc pcd_loc @@ fun () ->
           match pcd_args with
           | Pcstr_tuple(typs) ->
             exprn quoter typs |>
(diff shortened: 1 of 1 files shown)


src_plugins/ppx_deriving_eq.cppo.ml
```diff
@@ -148,7 +148,8 @@ let str_of_type ~options ~path ({ ptype_loc = loc } as type_decl) =
     | Ptype_abstract, Some manifest -> expr_of_typ quoter manifest
     | Ptype_variant constrs, _ ->
       let cases =
-        (constrs |> List.map (fun { pcd_name = { txt = name }; pcd_args } ->
+        (constrs |> List.map (fun { pcd_name = { txt = name }; pcd_args; pcd_loc } ->
+          (* with_default_loc pcd_loc @@ fun () -> *)
```
whitequark: Is this intended to be commented?

gasche: That was a debug thing, of course (to observe the previous state), will force-push. Sorry!

gasche:
Note that this only fixes the ppx-eq plugin that I happened to be testing with. There might be a more useful, general solution that would apply to all plugins, but I don't see it. (back in the deay, @yallop's deriving had a nifty library support that would handle that sort of things in a codebase shared between plugins, if I remember correctly.)

whitequark:
Great! ppx_deriving was long overdue for something like this and it annoyed me, but not enough to sit down and implement.

The global state is a bit unclean but I'm fine with it.

> (back in the deay, @yallop's deriving had a nifty library support that would handle that sort of things in a codebase shared between plugins, if I remember correctly.)

ppx_deriving is supposed to be this library, or at least was, but I already factored out everything that was meaningfully factorable out without turning ppx_deriving into some kind of obtuse framework that needs to have its tendrils deep in every plugin. And this is unobtrustive enough that all plugins can probably simply be amended...

PR #94, opened 2016-04-17
Support deriving from module type declarations.

paurkedal:
I could use an extension of the deriving interface to allow deriving definitions from module types.  The derived definitions would typically be a structure of the given signature or a functor producing or depending on it, though it could be regular functions involving first-class modules of the signature.  My current use-case is to generate the RPC client and server parts for a Thrift service.

### DIFF
--- a/src/ppx_deriving.cppo.ml
+++ b/src/ppx_deriving.cppo.ml
@@ -19,10 +19,16 @@ type deriver = {
                    type_declaration list -> structure;
   type_ext_str : options:(string * expression) list -> path:string list ->
                   type_extension -> structure;
+  module_type_decl_str : options:(string * expression) list ->
+                          path:string list ->
+                          module_type_declaration -> structure;
   type_decl_sig : options:(string * expression) list -> path:string list ->
                    type_declaration list -> signature;
   type_ext_sig : options:(string * expression) list -> path:string list ->
                   type_extension -> signature;
+  module_type_decl_sig : options:(string * expression) list ->
+                          path:string list ->
+                          module_type_declaration -> signature;
 }
 
 let registry : (string, deriver) Hashtbl.t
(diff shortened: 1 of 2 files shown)


paurkedal:
I noticed this breaks ppx_type_conv, since it registers `Ppx_deriving` with their own signature, see

* https://github.com/janestreet/ppx_type_conv/blob/master/ppx_deriving_exporter/ppx_type_conv_deriving.ml
* https://github.com/janestreet/ppx_type_conv/blob/master/src/type_conv.mli#L121-L135

It's documented as a private interface to be changed, so it may be a matter of synchronizing the releases.

src/ppx_deriving.cppo.ml
```diff
@@ -19,10 +19,14 @@ type deriver = {
                    type_declaration list -> structure;
   type_ext_str : options:(string * expression) list -> path:string list ->
                   type_extension -> structure;
+  modtype_str : options:(string * expression) list -> path:string list ->
```
whitequark: `module_type_decl`

whitequark:
@diml How do we properly do releases in lockstep?

diml:
I suppose instead of passing `(module Ppx_deriving)` I could expand all the functions, that will make the glue code more robust to small changes in ppx\_deriving. I'll add this change to the next minor release batch I'll do this week

whitequark:
Thanks!

paurkedal:
I pushed a version with modtype_* replaced by module_type_decl_*, I assume this is what you had in mind?

whitequark:
Yep--thanks!

diml:
ppx\_type\_conv 113.33.03 has the change I described

whitequark:
Ok--I will release a new ppx_deriving soon.

paurkedal:
Thanks to both of you!  (Not quite ready for a release myself.)

PR #98, opened 2016-05-02
Remove pervasives when appropriate.

Drup:
This removes the leading `Pervasives` module when matching on built-in types. It solves issues with ppx_import on 4.03 (where types are fully qualified).

### DIFF
--- a/src/ppx_deriving.cppo.ml
+++ b/src/ppx_deriving.cppo.ml
@@ -234,6 +234,33 @@ let attr_warning expr =
   let structure = {pstr_desc = Pstr_eval (expr, []); pstr_loc = loc} in
   {txt = "ocaml.warning"; loc}, PStr [structure]
 
+let attr_nobuiltin ~deriver attrs =
+  attrs |> attr ~deriver "nobuiltin" |> Arg.get_flag ~deriver
+let rec remove_pervasive_lid = function
+  | Lident _ as lid -> lid
+  | Ldot (Lident "Pervasives", s) -> Lident s
+  | Ldot (lid, s) -> Ldot (remove_pervasive_lid lid, s)
+  | Lapply (lid, lid2) ->
+    Lapply (remove_pervasive_lid lid, remove_pervasive_lid lid2)
+
+let remove_pervasives ~deriver typ =
+  if attr_nobuiltin ~deriver typ.ptyp_attributes then typ
+  else
+    let open Ast_mapper in
+    let map_typ mapper typ = match typ.ptyp_desc with
+      | Ptyp_constr (lid, l) ->
+        let lid = {lid with txt = remove_pervasive_lid lid.txt} in
+        {typ with
+         ptyp_desc = Ptyp_constr (lid, List.map (mapper.typ mapper) l)}
+      | Ptyp_class (lid, l) ->
+        let lid = {lid with txt = remove_pervasive_lid lid.txt} in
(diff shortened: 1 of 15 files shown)


src/ppx_deriving.cppo.ml
```diff
@@ -234,6 +234,33 @@ let attr_warning expr =
...
+let attr_nobuiltin ~deriver attrs =
+  attrs |> attr ~deriver "nobuiltin" |> Arg.get_flag ~deriver
+let rec remove_pervasive_lid = function
+  | Lident _ as lid -> lid
+  | Ldot (Lident "Pervasives", s) -> Lident s
+  | Ldot (lid, s) -> Ldot (remove_pervasive_lid lid, s)
+  | Lapply (lid, lid2) ->
+    Lapply (remove_pervasive_lid lid, remove_pervasive_lid lid2)
+
+let remove_pervasives ~deriver ty =
```
whitequark: `typ` please, to be consistent with code elsewhere.

src/ppx_deriving.cppo.ml
```diff
@@ -234,6 +234,33 @@ let attr_warning expr =
...
   {txt = "ocaml.warning"; loc}, PStr [structure]
 
+let attr_nobuiltin ~deriver attrs =
+  attrs |> attr ~deriver "nobuiltin" |> Arg.get_flag ~deriver
+let rec remove_pervasive_lid = function
+  | Lident _ as lid -> lid
+  | Ldot (Lident "Pervasives", s) -> Lident s
+  | Ldot (lid, s) -> Ldot (remove_pervasive_lid lid, s)
+  | Lapply (lid, lid2) ->
+    Lapply (remove_pervasive_lid lid, remove_pervasive_lid lid2)
```
whitequark: I think this isn't right. It does too much. `Foo.Pervasives.x` shouldn't be reduced to `Foo.x` or at all.

whitequark:
Also, any tests?

src/ppx_deriving.cppo.ml
Drup: it hits the `typ` used for the mapper later ... :/

src/ppx_deriving.cppo.ml
Drup: `Foo.Pervasives.x` is not reduced, but `Pervasives.Foo(Pervasives.M).t` is.

src/ppx_deriving.cppo.ml
whitequark: `map_typ` :)

Drup: Done.

src/ppx_deriving.cppo.ml
whitequark: But there are no modules in `Pervasives`.

Drup: Does it matter ? `Pervasives` is a toplevel name, so we reduce all path that starts with `Pervasives`, it's a simple specification.

whitequark: I suppose not.

Drup:
It's surprisingly annoying to test, since only ref (and result) is in pervasive. I added a test with show.

Drup:
I pushed more test and fixed deriving fold for ref.

whitequark:
Thanks

Issue #83, opened 2016-03-10
Type error with inheritied polymorphic variant type in map


andrewray:
Possibly an unusual case but if you declare the following types with a signature to constraint it

```
type 'a k = [ `A of 'a | `B of 'a k ] [@@deriving map,show]
type 'a l = [ 'a k | `C of 'a ] [@@deriving map,show]
```

you get the error

```
signature mismatch:
       ...
       Values do not match:
         val map_l :
           ('a -> 'b) ->
           [< `A of 'a | `B of 'a k | `C of 'a ] ->
           ([> `A of 'b | `B of 'c | `C of 'b ] as 'c)
       is not included in
         val map_l : ('a -> 'b) -> 'a l -> 'b l
       File "test.ml", line 21, characters 0-31: Expected declaration
       File "test.ml", line 75, characters 0-53: Actual declaration
```

I think the error is in the result `B of 'c` part, but I am not sure exactly how ocaml is inferring things here - apparently should B accept a C which seems tricky.

whitequark:
No, that seems right. I suspect the problem might be in the variance annotation, but not sure.

whitequark:
@andrewray Any chance you could submit a fix for this?

andrewray:
Having a look at this now.  In this bit of code

```
let rec expr_of_typ typ =
    ....
        | Rinherit ({ ptyp_desc = Ptyp_constr (tname, _) } as typ) ->
          Exp.case [%pat? [%p Pat.type_ tname] as x]
                   [%expr ([%e expr_of_typ typ] x :> ???)]
```

I want to replace the `???` with the actual type.  I dont see how to get this within `expr_of_typ` without changing it's signature (`core_type -> expression`) to something like `type_declaration -> core_type -> expression`.    However, this function also seems to be part of the Ppx_deriving.deriver structure so I am not too sure about changing this api.

whitequark:
Closed by #106.

PR #106, opened 2016-06-30
Fix for issue with inherited polymorphic variants

andrewray:
...and type sigs as per #83.

changes `expr_of_typ` to receive a type_declaration and hides it
behind an option type (note; possibly better to provide a `core_type
option` instead).

from within the deriver structure, using `expr_of_typ` to generate code
for inherited polymorphic variants will raise an error.  I don't know
if this is actually done anywhere.


whitequark:
Thanks!

PR #109, opened 2016-08-04
Add api for ppx_type_conv

rgrinberg:
This consists of two parts:
* A way to fetch all currently registered derivers
* A way to add hooks for when new derivers are registered

@diml is this good enough?

### DIFF
--- a/src/ppx_deriving.mli
+++ b/src/ppx_deriving.mli
@@ -42,6 +42,13 @@ type deriver = {
 (** [register deriver] registers [deriver] according to its [name] field. *)
 val register : deriver -> unit
 
+(** [add_register_hook hook] adds [hook] to be executed whenever a new deriver
+    is registered. *)
+val add_register_hook : (deriver -> unit) -> unit
+
+(** [derivers ()] returns all currently registered derivers. *)
+val derivers : unit -> deriver list
+
 (** Creating {!deriver} structure. *)
 val create :
   string ->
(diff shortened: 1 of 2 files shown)


diml:
Yes, that's enough to use ppx\_deriving plugins with ppx\_type\_conv. Note that `derivers` can probably just return a `deriver list` given that the key is the same as the `name` field

rgrinberg:
@diml sure. I made the fix.

src/ppx_deriving.mli
```diff
@@ -42,6 +42,10 @@ type deriver = {
 (** [register deriver] registers [deriver] according to its [name] field. *)
 val register : deriver -> unit
 
+val add_register_hook : (deriver -> unit) -> unit
```
whitequark: Documentation?

rgrinberg: alright

src/ppx_deriving.mli
```diff
@@ -42,6 +42,13 @@ type deriver = {
 (** [register deriver] registers [deriver] according to its [name] field. *)
 val register : deriver -> unit
 
+(** [add_register_hook hook] adds [hook] to be executed whenever a new deriver
+    is registered *)
+val add_register_hook : (deriver -> unit) -> unit
+
+(** [derivers ()] returns all currently registered derivers *)
```
whitequark: Nitpick: the sentences in the docs are finished with dots. I'll merge after this is fixed!

rgrinberg:
No problem.

PR #114, opened 2016-09-09
Fix variant formatting #46

### LINKED ISSUE #46: show: missing parentheses around constructors
Right now, only constructors with one argument have parenthesis around them. Omitting parenthesis is ok for nullary constructors, but shouldn't be for n-ary ones (as the the output can no longer simply be pasted into the top-level):

 ~~~OCaml
type t = A | B of t | C of t*t [@@deriving show];;
show @@ A;; (* A *)
show @@ B A;; (* (B A) *)
show @@ C (A, A);; (* C (A, A) but should be (C (A, A)) *)
show @@ B (C (A, A));; (* (B C (A, A)) <- problem *)
~~~

jaredly:
See issue for more info & discussion
fixes #46

### DIFF
--- a/src_plugins/ppx_deriving_show.cppo.ml
+++ b/src_plugins/ppx_deriving_show.cppo.ml
@@ -117,7 +117,7 @@ let rec expr_of_typ quoter typ =
           function
           | None -> Format.pp_print_string fmt "None"
           | Some x ->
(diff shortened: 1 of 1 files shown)


whitequark:
Bleh, you broke a test...

whitequark:
I've fixed that in master.

PR #115, opened 2016-09-22
Monomorphize compare function for builtin types

emillon:
The `[%ord: int]` expression expands to `Pervasives.compare`. Since this function is polymorphic, it may be used with other types. For example, the following expression typechecks:

```.ocaml
([%ord: int] : string -> string -> int)
```

This adds a type constraint on the `Pervasives.compare` call so that it becomes monomorphic. This is also more consistent with how this case is handled in the `eq` plugin.

Let me know what you think. Thanks!

### DIFF
--- a/src_plugins/ppx_deriving_ord.cppo.ml
+++ b/src_plugins/ppx_deriving_ord.cppo.ml
@@ -75,7 +75,8 @@ and expr_of_typ quoter typ =
               | [%type: int64] | [%type: Int64.t] | [%type: nativeint]
               | [%type: Nativeint.t] | [%type: float] | [%type: bool]
               | [%type: char] | [%type: string] | [%type: bytes]) ->
-        [%expr Pervasives.compare]
+        let compare_fn = [%expr fun (a:[%t typ]) b -> Pervasives.compare a b] in
+        Ppx_deriving.quote quoter compare_fn
       | true, [%type: [%t? typ] ref] ->
         [%expr fun a b -> [%e expr_of_typ typ] !a !b]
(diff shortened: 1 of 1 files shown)


emillon:
Sorry, I was not invoking the test suite the correct way.

The problem with my patch is that in the following generated code:

```.ocaml
type es =
  | ESBool of bool 
  | ESString of string 
and bool =
  | Bfoo of int* ((int -> int)[@compare fun _  -> fun _  -> 0]) 
and string =
  | Sfoo of String.t* ((int -> int)[@compare fun _  -> fun _  -> 0]) 
[@@deriving ord]
let rec (compare_es : es -> es -> Ppx_deriving_runtime.int) =
  ((let open! Ppx_deriving_runtime in
      fun lhs  ->
        fun rhs  ->
          match (lhs, rhs) with
          | (ESBool lhs0,ESBool rhs0) ->
              ((fun (a : bool)  -> fun b  -> Pervasives.compare a b)) lhs0 rhs0
```

`bool` resolves to `Ppx_deriving_runtime.bool` which is not the correct one. I am not too familiar with how quoting works in this project, so I am not sure how to fix that. Is it possible to quote types and not expressions?

whitequark:
Your tests fail on Travis:
```
- File "src_test/test_deriving_ord.cppo.ml", line 123, characters 0-211:
- Error: This expression has type bool/2260
-        but an expression was expected of type
-          Ppx_deriving_runtime.bool = bool/5
```

emillon:
Just fixed the tests (and quoting problems) by adding a call to `Ppx_deriving.quote`.

Issue #116, opened 2016-11-07
type nonrec is ignored causing an infinite loop


emillon:
Hi,

I'm trying to use is with `type nonrec` but it seems that this annotation is ignored.

Example:

```ocaml
type t = int
[@@deriving ord]

module M = struct
  type nonrec t = t
  [@@deriving ord]
end

let () = Printf.printf "%d\n" (M.compare 1 2)
```

This triggers an infinite loop instead of calling the top-level `compare` function.

Thanks for this great project!

whitequark:
Please note that I do not have much time to maintain this library, and will not fix this bug. I can offer you assistance in writing a PR though.

emillon:
Sure, I'd be interested. Any pointers would be appreciated.

whitequark:
@emillon Sure. First, try compiling your code using `ocamlc -dsource`. This will output the preprocessed source as currently used. I think you will quickly spot why implementing `type nonrec` is problematic; this messes rather badly with the hygiene mechanism. (Also try using `[@@deriving]` with a let..and phrase to see the source of additional complexity.)

whitequark:
I think rewriting the output of ppx_deriving such that it works with type nonrec will already go a long way towards fixing this.

XVilka:
This issue is mentioned in Facebook's Infer [Contributors Guide](https://github.com/facebook/infer/blob/master/CONTRIBUTING.md#ocaml).

gasche:
Recent versions of ppx_deriving (4.1.5 (released in November 2017) and later) show an error when `nonrec` is used, so at least there is no silent production of looping code anymore.

I had a look last week and my impression is that it is not an easy problem to solve. At first it looks like just not adding the `rec` keyword would make the resulting code sensible for a `nonrec` type (the name of the generated function would refer to the generated function of the previous type of the same name, which is what we want), but in fact the structure of the returned code is decided by the plugins, not by the API module (which mostly provide helpers to make plugin life easier), so (1) fixing this issue would require a change in each of the plugins and (2) we have to do some API design/changes if we want to push more logic into the API module. Then I ran out of time for the day, and I stopped looking at it.

Remark: some plugins have evolved in a way that uses the recursivity of the generator for convenience; for example `show` generates a `show` and `pp` function, the definition of `show` depends on `pp`, but there is no explicit sequencing of declarations as they are in a big mutually-recursive block. To stop being recursive by default, we have to break those mutual blocks in the plugins.

PR #118, opened 2016-12-21
Raise an error when `type nonrec` is encountered

emillon:
This feature has two different implementations in the AST:

- in 4.02.3, it is an annotation on a type,
- in >= 4.03.0 it is straight in the constructor.

Note that this can break existing programs; however relying on `ppx_deriving` to do the right thing in that case seems dangerous.
Maybe an escape hatch could be added so that the recursive version is derived nonetheless.

It is an intermediate solution to #116. Solving it is tricky because it would be necessary to expose this to the plugins, breaking the API. It is probably easier to do so once 4.02.3 support is dropped, since the per-type semantics are difficult to handle.

Let me know what you think about these changes.
Thanks!

### DIFF
--- a/src/ppx_deriving.cppo.ml
+++ b/src/ppx_deriving.cppo.ml
@@ -462,6 +462,21 @@ let module_from_input_name () =
   | "//toplevel//" -> []
   | filename -> [String.capitalize (Filename.(basename (chop_suffix filename ".ml")))]
 
+let pstr_desc_rec_flag pstr =
+  match pstr with
+  | Pstr_type(rec_flag, typ_decls) ->
+#if OCAML_VERSION < (4, 03, 0)
+    begin
+      if List.exists (fun ty -> has_attr "nonrec" ty.ptype_attributes) typ_decls then
+        Nonrecursive
+      else
+        Recursive
+    end
+#else
+    rec_flag
+#endif
+  | _ -> assert false
+
 let mapper =
   let module_nesting = ref [] in
   let with_module name f =
@@ -496,6 +511,10 @@ let mapper =
   in
   let structure mapper items =
     match items with
+    | { pstr_desc = Pstr_type(_, typ_decls) as pstr_desc ; pstr_loc } :: rest when
+        List.exists (fun ty -> has_attr "deriving" ty.ptype_attributes) typ_decls
+        && pstr_desc_rec_flag pstr_desc = Nonrecursive ->
+      raise_errorf ~loc:pstr_loc "The nonrec flag is not supported by ppx_deriving"
     | { pstr_desc = Pstr_type(_, typ_decls); pstr_loc } as item :: rest when
         List.exists (fun ty -> has_attr "deriving" ty.ptype_attributes) typ_decls ->
       let derived =


whitequark:
> Solving it is tricky because it would be necessary to expose this to the plugins, breaking the API. It is probably easier to do so once 4.02.3 support is dropped, since the per-type semantics are difficult to handle.

What API do you propose?

emillon:
I was thinking about extending the type of the `type_decl_str` argument (now `options:(string * expression) list -> path:string list -> type_declaration list -> structure`) so that it takes an extra `recursive:bool` argument.

Alternatively, maybe this could be encoded in the `options` list (with `ppx_deriving` adding something like `("ppx_deriving.recursive", [%expr true])`.

So far I haven't tried to make a plugin nonrec-aware, so I'm not sure about the implications on that side.

whitequark:
> I was thinking about extending the type of the type_decl_str argument (now options:(string * expression) list -> path:string list -> type_declaration list -> structure) so that it takes an extra recursive:bool argument.

How about adding a `?recursive:bool`? This should preserve source compatibility.

emillon:
Absolutely, that would probably work too.

whitequark:
@emillon poke? Any chance you could implement this?

emillon:
Hi,

I had a look at this, and there are two challenges:
- it's a bit tricky to add an optional argument, because of polymorphism in functions such as  `Ppx_deriving.derive`. It might be possible to use some partial application tricks which are not pretty, but it's probably worth preserving source compatibility as you said.
- there are visibility issues in some cases. At first I thought that the rec_flag for the value declaration should be the same as the type declaration, but that's not the case: for example in `show`, it seems that `pp` and `show` should be mutually recursive.

These do not seem to be show stoppers but this is a bit more involved than I initially thought.

whitequark:
Thanks for the analysis. I would really appreciate any implementation, yours or by anyone else.

gasche:
Regarding code generation, the scheme I would propose is to turn

```ocaml
let rec fn_foo = ...
and fn_bar = ...
```

into

```
let rec fn_foo_nonrec = ...
and fn_bar_nonrec = ...
let fn_foo = fn_foo_nonrec
and fn_bar = fn_bar_nonrec
```

or a variant of this that hides the `_nonrec` declaration from the global scope, such as

    let fn_foo, fn_bar =
      let rec fn_foo_nonrec = ...
      and fn_bar_nonrec = ....
    in fn_foo_nonrec, fn_bar_nonrec

Note in particular that occurrences of `foo`, `bar` in the declaration will lookup a `fn_{foo,bar}` in the environment, as expected, while mutual recursion of the two functions is still possible.

It is not immediate to implement this, and I think there is no solution that is both good and backward-compatible with the previous API. (One reason in particular is that currently it is the plugin's duty to generate the whole str-item, so this has to be changed plugin-wise and we need to pass them more information as @emillon suggest).
I would personally give up on backward-compatibility: plugins can easily use version bounds to advertise the version of ppx_deriving they support, allowing you to keep a clean API close to the current parsetree. One could encapsulate the logic of this code-generation scheme (which is not simple) in a Ppx_deriving function, so that user plugins can be updated by just calling this function in their `type_decl_str` component -- `type_decl_sig` does not change, as the external interface is not affected by `nonrec`.

whitequark:
I agree on giving up backwards compatibility.

PR #120, opened 2017-02-09
Show plugin: add option to skip module path

Kakadu:
We were kind of bothered that `show` adds filename as prefix for all constructors. So we added very straightforward way to workaround this.  What was the expected way to hack this?

@eucpp

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,24 @@ It is also possible to use `[@polyprinter]`. The difference is that for a type `
 
 The function `fprintf` is locally defined in the printer.
 
+By default all constructors are printed with prefix which is dot-separated filename and module path. For example
+``` ocaml
+# module X = struct type t = C [@@deriving show] end;;
+...
+# X.(show C);;
+- : Ppx_deriving_runtime.string = "X.C"
+```
+
+This code will create printers which return the string `X.C`, `X` is a module path and `C` is a constructor name. File's name is omitted in the toplevel. To skip all module paths the one needs to derive show with option `with_path` (which defaults to `true`)
+
+``` ocaml
+# module X = struct type t = C [@@deriving show { with_path = false }] end;;
+...
+# X.(show C);;
+- : Ppx_deriving_runtime.string = "C"
+```
+
+
 Plugins: eq and ord
 -------------------
 
--- a/src_plugins/ppx_deriving_show.cppo.ml
+++ b/src_plugins/ppx_deriving_show.cppo.ml
@@ -12,10 +12,15 @@ open Ast_convenience
 let deriver = "show"
 let raise_errorf = Ppx_deriving.raise_errorf
 
+type options = { with_path : bool }
+
 let parse_options options =
+  let with_path = ref true in
   options |> List.iter (fun (name, expr) ->
     match name with
-    | _ -> raise_errorf ~loc:expr.pexp_loc "%s does not support option %s" deriver name)
+    | "with_path" -> with_path := Ppx_deriving.Arg.(get_expr ~deriver bool) expr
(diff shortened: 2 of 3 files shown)


whitequark:
This is implemented through global state and is completely unacceptable. Instead, move the state to the ~type_decl_str function.

Kakadu:
Better?

hcarty:
@Kakadu Could you add a doc update to go with it?  This looks like a useful feature.

Kakadu:
@hcarty Are you talking about `README.md `?

tekknolagi:
@Kakadu that seems reasonable. As someone who is currently trying to use this API and would appreciate this option, docs anywhere would be helpful.

hcarty:
@Kakadu Yes, to echo @tekknolagi.  An update to `README.md` would be useful.

Kakadu:
@hcarty @tekknolagi  ^^

README.md
```diff
@@ -153,6 +153,22 @@ It is also possible to use `[@polyprinter]`. The difference is that for a type `
 
 The function `fprintf` is locally defined in the printer.
 
+By default all constructors are printed with prefix which is dot-separated filename and module path. For example
+``` ocaml
+(* A.ml *)
+module X = struct type t = C [@@deriving show] end
+```
+
+This code will create printers which return the string `A.X.C` where `A` is a filename, `X` is a module path and `C` is a constructor name. To skip all module paths the one needs to derive show with `nofullpath` option (which equals to `false` by default)
```
hcarty: "which defaults to `false`" would be more clear here.

README.md
```diff
@@ -153,6 +153,22 @@ It is also possible to use `[@polyprinter]`. The difference is that for a type `
...
+
+``` ocaml
+# module X = struct type t = C [@@deriving show { nofullpath = true }] end;;
+...
+# X.show (X.C);;
+- : Ppx_deriving_runtime.string = "C"
+```
```
hcarty: The first example shows just the code while this one is in the context of a toplevel session. Maybe they should both be in the toplevel or both be just code?

src_plugins/ppx_deriving_show.cppo.ml
```diff
@@ -12,10 +12,15 @@ open Ast_convenience
 let deriver = "show"
 let raise_errorf = Ppx_deriving.raise_errorf
 
+type options_t = { nofullpath : bool }
```
whitequark: Just `type options`.

src_plugins/ppx_deriving_show.cppo.ml
```diff
@@ -12,10 +12,15 @@ open Ast_convenience
...
+
 let parse_options options =
+  let nofullpath = ref false in
   options |> List.iter (fun (name, expr) ->
     match name with
-    | _ -> raise_errorf ~loc:expr.pexp_loc "%s does not support option %s" deriver name)
+    | "nofullpath" -> nofullpath := Ppx_deriving.Arg.(get_expr ~deriver bool) expr
```
whitequark: I'd prefer to avoid negation here. So, the option would be called `with_path` and default to `true`.

README.md
tekknolagi: What about using the API manually, as in a hand-written mapper?

Kakadu: Yeah, this change can be useful but I'm not psychologically ready to invest more time in it today.

Kakadu:
@whitequark Can you review again?

README.md
tekknolagi: @Kakadu will you before this PR is in?

Kakadu: I don't think so

PR #128, opened 2017-04-05
Support result and not just Result.result

bbc2:
On OCaml 4.03.0+, this enables you to write
```ocaml
[%show: (string, string) result] (Ok "hello")
```
instead of being limited to:
```ocaml
[%show: (string, string) Result.result] (Ok "hello")
```
This should work for all standard plugins, not just "show".


whitequark:
It seems mildly weird to accept just `result` even on pre-4.03, but I think that does not have any potential for harm, so no need to change anything.

Issue #130, opened 2017-04-10
Versioned API


diml:
Following the discussion in https://github.com/whitequark/ppx_deriving_yojson/pull/49, I'd like to look a adding some kind of versioned API to ppx_deriving. But I'd like to discuss the final picture before starting.

# Current state

At the tip of the master branch, `Ppx_deriving` is using the 4.05 Ast from ocaml-migrate-parsetree explicitly. This means that it breaks all packages using ppx_deriving, except on 4.05. I guess this is a bit abrupt?

# Proposal

## Initial setup

We restore the `Ppx_deriving` as it was before #123, i.e. using the Ast from the compiler with some cppo to keep it working with several versions of the compiler. However, we deprecate it. We also keep the registration with the omp driver.

The `Ppx_deriving` module as it is at the tip of master is copied to `Ppx_deriving_405` and distributed as `ppx_deriving.405`. At the same time, we take the opportunity to upgrade the API:

- add a `rec_flag` argument to the `type_decl_{str,sig}` callbacks
- replace `string` by `string Location.loc` in various places where a location was added

## Gluing thing together

At this point we have two independent rewriters interpreting `[@@deriving]` and `[%derive.foo]`. To solve this problem, we can reuse the idea of `register_external`: we add a module `Ppx_deriving_attribute` whose only function is to register the various derivers implemented by the various passes. Each pass will ignore the derivers that are interpreted by another pass.

## Upgrade story

In order to upgrade the version of the AST used by ppx_deriving, for instance from 4.05 to 4.07, one proceeds as follow:

- copy `Ppx_deriving_405` to `Ppx_deriving_407`
- deprecate `Ppx_deriving_405`
- upgrade `Ppx_deriving_407`

## Rationale

It the end it is better if ppx rewriters follow new versions of the Ast, since omp won't support all versions forever. With this plan, people have a window of time in which they can upgrade their code, without having to upgrade all ppx rewriters once.

## Tweak

If we do exactly as described, it seems that we'll regularly loose the history of the `Ppx_deriving_XXX` module since there is no `git cp`. To workaround this, we can keep the latest version of the code in `Ppx_deriving_latest` and then setup a rule to copy it to `Ppx_deriving_XXX` or use `include `Ppx_deriving_latest` in `Ppx_deriving_XXX`.

whitequark:
[...]

PR #132, opened 2017-04-20
Use ppx_derivers as registry

diml:
This implement the sub-request in #130.

Essentially what it does is move the registry into a separate [ppx_derivers package](
https://github.com/diml/ppx_derivers), which has a very simple API:

```ocaml
(** Ppx derivers

    This module holds the various derivers registered by either ppx_deriving or
    ppx_type_conv.
*)

(** Type of a deriver. The concrete constructors are added by
    ppx_type_conv/ppx_deriving. *)
type deriver = ..

(** [register name deriver] registers a new deriver. Raises if [name] is already
    registered. *)
val register : string -> deriver -> unit

(** Lookup a previously registered deriver *)
val lookup : string -> deriver option

(** [derivers ()] returns all currently registered derivers. *)
val derivers : unit -> (string * deriver) list
```

Since this supersede `register_external` which is not yet released, I just removed this function.

This patch makes things simpler for ppx_type_conv, since I can replace the optional dependency on ppx_deriving by a hard dependency on ppx_derivers.

### DIFF
--- a/_tags
+++ b/_tags
@@ -2,7 +2,7 @@ true: warn(@5@8@10@11@12@14@23@24@26@29@40), bin_annot, safe_string, debug, cppo
 "data": -traverse
 
 "src": include
-<src/*.{ml,mli,byte,native}>: package(dynlink ocaml-migrate-parsetree ppx_tools_versioned.metaquot_405 result)
+<src/*.{ml,mli,byte,native}>: package(dynlink ocaml-migrate-parsetree ppx_tools_versioned.metaquot_405 result ppx_derivers)
 <src/ppx_deriving_main.{ml,mli,byte,native}>: package(findlib.dynload)
 <src/ppx_deriving_main.{byte,native}>: predicate(ppx_driver), linkall
 <src_plugins/*.{ml,mli}>: package(ocaml-migrate-parsetree ocaml-migrate-parsetree ppx_tools_versioned.metaquot_405)
--- a/pkg/META.in
+++ b/pkg/META.in
@@ -15,7 +15,7 @@ package "runtime" (
 package "api" (
   version = "%{version}%"
   description = "Plugin API for ppx_deriving"
-  requires = "dynlink compiler-libs.common ppx_tools_versioned result"
+  requires = "dynlink compiler-libs.common ppx_tools_versioned result ppx_derivers"
   archive(byte) = "ppx_deriving.cma"
   archive(native) = "ppx_deriving.cmxa"
   exists_if = "ppx_deriving.cma"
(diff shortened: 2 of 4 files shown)


whitequark:
I still don't understand how is this supposed to work. The values of `deriver` not declared inside ppx_deriving are completely abstract to it. How would they be used in ppx_deriving?

diml:
Note that all the work was done in #126, this simply moves the  hash table outside of `Ppx_deriving`.

The idea is that for any deriver named `name` in the input, from the point of view of `Ppx_deriving` there are three possibilities:

- `Ppx_derivers.lookup name = Some (Ppx_deriving.T d)`: `name` was registered using `Ppx_deriving.register` and must be expanded by `Ppx_deriving`
- `Ppx_derivers.lookup name = Some <some other constructor>`: `name` was registered by another pass and will be expanded by this pass, it can be safely ignored by `Ppx_deriving`
- `Ppx_derivers.lookup name = None`: `name` is unknown, it is an error

And it is the same story for ppx_type_conv.

whitequark:
@diml OK. I now understand the architecture you propose. I have no objections to it, and will merge after a final look shortly.

Drup:
@diml Is there a reason not to make `ppx_derivers` as part of omp's driver ? It feels a bit redundant to have a 10 line package when it seems clearly related to another one.

diml:
It seems to me that the omp driver is common to all ppx rewriters while ppx_derivers is common to `[@@deriving ...]` plugins only

Drup:
Sure, but it's literally a 10 line piece of code that seems quite related to ppx driving. Derivers will already use omp, so ....

diml:
TBH, I don't see the problem with the package being small.

diml:
I've released ppx_derivers in opam: https://github.com/ocaml/opam-repository/pull/9058

whitequark:
Sorry for delaying this for so long--I did take the final look I promised, I have no objections and will merge and release ASAP since this doesn't alter public API.

diml:
A new release would be good. This should probably be backported to v4.1 since master breaks the API. I did this for our internal copy of ppx_deriving at Jane Street and added this at the end of `ppx_deriving.cppo.ml`:

```
(* This is only used when ppx_deriving is linked as part of an ocaml-migrate-parsetre
   driver. *)
let () =
  Migrate_parsetree.Driver.register ~name:"ppx_deriving"
    (module Migrate_parsetree.OCaml_current)
    (fun _ _ -> mapper)
```

This + compiling the plugins with -linkall is enough to make ppx_deriving work with the ocaml-migrate-parsetree driver, and so jbuilder, without changing things for existing users of ppx_deriving.

I created this branch to backport changes from master + this PR and the omp integration: https://github.com/diml/ppx_deriving/commits/v4

whitequark:
@diml is this still necessary?

whitequark:
Merged as 304a4f91, a part of #141.

Issue #153, opened 2017-10-14
understanding the current mess


gasche:
Hi all,

(cc @let-def, @trefis, @bikalgurung, @avsm, @xclerc, @alainfrisch )

It seems that the current state of ppx-deriving combined with the rest of the ppx ecosystem is fairly unsatisfying for now, as this repository is full of users for which the tool does not work any more: #143, #147, #148, #149, #150 are all issues that arose since the last release of ppx-deriving (4.2) with ocaml-migrate-parsetree (omp for short) integration. (Some deriving plugins are also affected, see https://github.com/ocaml-ppx/ppx_deriving_yojson/issues/56 , https://github.com/ocaml-ppx/ppx_deriving_yojson/issues/63. )

This integration was not painless:
- the previous ppx_deriving maintainer, @whitequark, has distanced themselves from maintenance of the library for many reasons and is not currently able to work more on improving the library (that's perfectly understandable and we should be able to get it to move forward on our own)
- @let-def, who created omp, proposed a first PR to use it for ppx-deriving (to get 4.05 compatibility "for free") as #123, which required a change in the way ppx_deriving is invoked
- @diml (who is currently in holidays and should thus not be expected to work on this issue) later proposed to revert this change and propose a "more compatible" approach, #141, which was eventually merged and is what remains in the 4.2 release.

I am not a heavy user of ppx-deriving myself, and far from an expert on ppx matters. My understanding of the issue is therefore almost non-existent, and I would be happy if other people took care of the issue for the 4.05 and 4.06 release cycles. However, progress has been slow to nonexistent so far, so I'm creating this issue to try to at least clarify the situation and understand what the problem is.

I have two separate questions:

1. There are plenty of examples of the fact that ppx-deriving currently does *not* work for many people. But who are the people for which it works? What are their development environment like? Does the current version of ppx-deriving work as expected for all `jbuilder` users, for example? Does it work for all people inside Jane Street (possibly using non-publicly-released software versions), but not outside it?

2. Is there a *simple* and *workable* way to get ppx-deriving back to "work as expected for OCaml 4.05 and 4.06, even if it is not the best long-term way to do things"? For example, would reverting the use of ocaml-migrate-parsetree and going back to cppo IFDEFS solve the problem? (Would it work for some of the current or prospective users of ppx-deriving, but not the others?).

[...]

PR #141, opened 2017-06-07
Revert #123 + omp driver integration

diml:
This patch does the following:

- it reverts #123 (sorry @let-def :()
- it includes #132
- it adds what is necessary to make ppx_deriving plugins work with the ocaml-migrate-parsetree driver
- it fixes the build with 4.05 using one more `#if`

The aim is to get a version of ppx_deriving that doesn't break compatibility and works with the ocaml-migrate-parsetree driver, and so jbuilder.

This code is what we have been using inside Jane Street for a while as well as what is used by our development opam repository, so it is tested. So far everything seems to work as expected:
- using ppx_deriving and/or ppx_type_conv plugins via `ocamlfind -package ...` still works as before
- using ppx_deriving and/or ppx_type_conv plugins with the ocaml-migrate-parsetree driver, either with the ocamlbuild plugin or with jbuilder now works

I'm interested to see this merged and released as it will allow to use ppx_deriving plugins with jbuilder without hassle.

BTW, the diff is much smaller if you read it after the revert of #123.

PR #154, opened 2017-10-24
make ppx_deriving 4.1 build under 4.05

gasche:
See #153: this is the start of an attempt to provide a 4.1 maintenance release that supports recent OCaml versions -- as a temporary solution to the current breakage for non-driverized users.

I started by creating a branch [4.1-maintenance](https://github.com/ocaml-ppx/ppx_deriving/commits/4.1-maintenance) on the upstream repo, which contains all the commits that went into 4.2 except for the omp-support changes in #141 -- currently it points exactly to 6b75b9e4804b253945ab61a76ed4af539d6caf90. A first point of potential discussion is whether we should indeed start from all those commits (some of there are small feature requests rather than bugfixes), but I think this is a reasonable basis to start from.

Then I just implemented 4.05-support using ugly preprocessor conditionals. This is the current PR, and it is very simple. I also implemented 4.06-support (which is a bit heavier due, I think, to oversight in the parsetree evolution), which will come as a separate PR.

If there is consensus to merge this PR, this means that users of 4.1-style ppx_deriving will be able to use it under 4.05.0.  
It is an explicit design goal of mine that users of 4.2-style ppx_deriving (in particular jbuilder users) would *not* be affected by a release of a 4.1.1 or 4.1+4.05 maintenance release, but I have not tested for it -- this will have to be tested at the time of (maybe) publishing a 4.1.foo on opam, not from the current PR, but if/when this happens the testing help of jbuilder users will be warmly welcome.

### DIFF
--- a/opam
+++ b/opam
@@ -25,6 +25,7 @@ depends: [
   "ocamlbuild" {build}
   "ocamlfind"  {build & >= "1.6.0"}
   "cppo"       {build}
+  "cppo_ocamlbuild" {build}
   "ppx_tools"  {>= "4.02.3"}
   "result"
   "ounit"      {test}
--- a/src/ppx_deriving.cppo.ml
+++ b/src/ppx_deriving.cppo.ml
@@ -327,6 +327,9 @@ let free_vars_in_core_type typ =
       List.map free_in xs |> List.concat
     | { ptyp_desc = Ptyp_alias (x, name) } -> [name] @ free_in x
     | { ptyp_desc = Ptyp_poly (bound, x) } ->
+#if OCAML_VERSION >= (4, 05, 0)
+      let bound = List.map (fun y -> y.txt) bound in
+#endif
       List.filter (fun y -> not (List.mem y bound)) (free_in x)
     | { ptyp_desc = Ptyp_variant (rows, _, _) } ->
       List.map (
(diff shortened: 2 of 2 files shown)


gasche:
(One thing that is missing from the maintenance branch is the wider coverage by travis of OCaml versions. I will send a separate pull-request to add this.)

rleonid:
@gasche I'm sure that you're busy, but is there something that is blocking this PR?

I'm trying to fix the versions of the code that I'm using in a program.

gasche:
@rleonid I was hoping for some sort of feedback by someone else, but then I decided that I would merge anyway this week -- and also consider making a release.

I see now that Travis is failing for the missing cppo_ocamlbuild dependency, we also need to fix this first.

rleonid:
My feedback is a big 👍  , does that count?

I think that this completes the first 2 points of @avsm 's requests in the issue-that-shall-be-named, and having a separate, non-master, branch is a good way to proceed. Can we do something similar for `ppx-deriving-yojson` ?

gasche:
That's helpful, thanks. Should I look at ppx-deriving-yojson, or would you for example send a similar PR there?

rleonid:
I've already created a [pr](https://github.com/ocaml-ppx/ppx_deriving_yojson/pull/64), but maybe there should be a similar 4.1-maintenance branch, and the PR can go there instead of master. I guess, one can the branch for my PR and rename it?

gasche:
Sorry, of course I've seen this PR. If I understand correctly, there isn't anything to "revert" before continuing the temporary cppo-approch in ppx_deriving_json, so I think going on top of master is fine. That said, I do have branch-creation rights over the ocaml-ppx universe, so I would be happy to humor you if you do think that working in a separate branch would be better.

rleonid:
Oh true. No need to humor me.

gasche:
Note: I want to investigate making changes to this PR as suggested by https://github.com/ocaml-ppx/ppx_deriving_yojson/pull/64.

gasche:
As a result of the investigation, I propose to merge https://github.com/ocaml-ppx/ppx_deriving/pull/159, which is an API change designed to minimize client-side changes (the API changes plays nicer with Parsetree changes foon the same OCaml versions).

PR #155, opened 2017-10-24
4.06 support

gasche:
This PR sits on top of #154. It adds support for 4.06.0 -- see also #153.

The changes in the drivers are very repetitive and annoying. It comes from the fact that the `label` field (morally) of the `Rtag` parsetree constructor gained a location, but the `label` fields of pattern and expression variant constructors is not located. If I had done this work earlier in the 4.06 release timeline, I would push for changing the parsetree to locate those labels as well -- this is clearly what the code tell us is the right approach for consumers. But I think this is too late for a fix in 4.06, so this will have to wait for 4.07 or never -- and the kludgy conditionals will remain, oh well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,18 @@
 language: c
-script:
-  - echo "yes" | sudo add-apt-repository ppa:avsm/ocaml42+opam12
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq opam ocaml-native-compilers
-  - export OPAMYES=1
-  - opam init
-  - eval `opam config env`
-  - opam pin add -n ppx_tools git://github.com/alainfrisch/ppx_tools#4.02
-  - opam pin add -n -k path ppx_deriving .
-  - opam install --deps-only -d -t ppx_deriving
-  - opam install -d -t -v ppx_deriving
+sudo: false
+services:
+  - docker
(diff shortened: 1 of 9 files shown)


gasche:
This PR also needs a 4.06-supporting version of ppx_tools, which does not currently exist. I created one locally and am in the process of submitting it upstream in https://github.com/ocaml-ppx/ppx_tools/pull/62.

gasche:
cc @alainfrisch : you may have an opinion on the located-label issue I mention above.

copy:
I believe this PR also needs to be applied on the master branch to provide 4.06 support for jbuilder users.

hannesm:
AFAICT, there is no `ppx_deriving` in the opam-repository with 4.06 support.  Is there any chance we can get that soon?  it looks like the code is there, and it "just" needs someone tagging a release and opening a PR on the opam-repository!?

gasche:
There is an interface-changing change to the 4.05 support that I want to consider before a release. I haven't had any time to work on it in the last few days, but I hope to get it done tomorrow (or decide not to do it), after which I will tag the release and send a PR.

copy:
Can we also consider #156 before a release?

gasche:
My plan, I think, was to make a `4.1+` release. I think your PR, if/when merged, would be part of a `4.2+` release.

PR #156, opened 2017-11-05
4.06 support for ppx_deriving 4.2

copy:
This is a cherry-pick of @gasche changes from #155 to make the driverized version in the master branch work on OCaml 4.06.

This fixes ppx_deriving for jbuilder users when using ppx_deriving with other driverized ppxs. The error message was `Error: ppx_type_conv: 'show' is not a supported type type-conv generator`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
   - PACKAGE="ppx_deriving"
   - TESTS="true"
   matrix:
+  - DISTRO="alpine" OCAML_VERSION="4.06.0"
   - DISTRO="alpine" OCAML_VERSION="4.05.0"
   - DISTRO="alpine" OCAML_VERSION="4.04.2"
   - DISTRO="alpine" OCAML_VERSION="4.04.1"
--- a/opam
+++ b/opam
@@ -25,6 +25,7 @@ depends: [
   "ocamlbuild" {build}
   "ocamlfind"  {build & >= "1.6.0"}
   "cppo"       {build}
+  "cppo_ocamlbuild" {build}
   "ocaml-migrate-parsetree"
   "ppx_derivers"
   "ppx_tools"  {>= "4.02.3"}
--- a/src/ppx_deriving.cppo.ml
+++ b/src/ppx_deriving.cppo.ml
@@ -12,6 +12,12 @@ open Parsetree
 open Ast_helper
 open Ast_convenience
 
+#if OCAML_VERSION >= (4, 05, 0)
+type tyvar = string Location.loc
+#else
+type tyvar = string
+#endif
+
 type deriver = {
   name : string ;
   core_type : (core_type -> expression) option;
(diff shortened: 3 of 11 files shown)


gasche:
I find it surprising to add conditional-compilation directives to ppx_deriving 4.2, as my understanding was that the 4.2 code structure used ocaml-migrate-parsetree instead of conditional compilation. Are there still existing cppo directives in the 4.2 codebase that I didn't know about, or is your PR re-introducing conditionals that didn't exist before? Is there a solution to 4.06 support using ocaml-migrate-parsetree without conditionals?

copy:
The ocaml-migrate-parsetree changes were reverted in #123, after which omp was only used to register this ppx for jbuilder. The rationale was discussed in https://github.com/ocaml-ppx/ppx_deriving/issues/130.

copy:
By the way, this branch also works for ocamlbuild users (and I assume users of other build tools as well). Only adding JaneStreet's ppxs to the mix seems to cause problems.

gasche:
Update: I worked more on 4.05 support in ppx_deriving (inspired by the ppx_deriving_yojson work) and hope to send a PR tonight. Then the current PR would have to be adapted (add more commits).

Independently, I convinced myself that trusting @copy here cannot hurt: the code is not worse with the PR merged than it was before, and I'm happy to have his testing contrbutions. So my plan is to merge the PR (when it converged with my 4.05-level changes), and cut a 4.2 maintenance release after that -- if @copy agrees.

copy:
> So my plan is to merge the PR (when it converged with my 4.05-level changes), and cut a 4.2 maintenance release after that -- if @copy agrees.

Sounds good to me. I've updated the PR accordingly.

gasche:
Thanks! Merged.

As time passes I feel a bit less confident about the 4.1-maintenance branch. There are two things:
1. You seem to say that 4.2 works fine for many users, and I have been told that some of the driver-side issues (independent from ppx_deriving) have been fixed or are being fixed.
2. I have second-thoughts about the fact that I merged in the 4.1-maintenance branch some things that are *features* developed before the 4.2 release. The idea at the time was to "cut before the weird stuff", but I am uncomfortable tagging 4.1.2 (or 4.1.5) something that is not just a bugfix release.

I am hesitating between dropping the 4.1-maintenance idea completely, or releasing the branch as-is (4.1.5), or removing the new features that I cherry-picked in it to do an actual minor release (4.1.2).

gasche:
(My short-term branch is to test `master` on my machine and see whether I can at least build the external plugins from it.)

gasche:
I tested the current `master` on my machine, in the sense that I used it to run the `make test` targets of the corresponding ppx_deriving projects (protobuf, yojson) and they seems to work just fine. I will make a maintenance release of 4.2, and wait to see feedback from user to know whether I still need a 4.1 release as well.

PR #165, opened 2018-03-06
Use Caml.Format in show, to avoid warnings in Base

stevejb71:
Jane Street's Base emits a warning if Format is used without
the Caml. qualifier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+NEXT_RELEASE
+------------
+
+* use Format through Ppx_deriving_runtime to avoid deprecation warning
+  for users of JaneStreet Base
+  (Stephen Bastians and Gabriel Scherer, review by whitequark)
+* silence a ambiguous-field warning (41) in generated code
+  #163
+  (Étienne Millon, review by Gabriel Scherer)
+
 4.2.1
 -----
 
--- a/src/ppx_deriving_runtime.mli
+++ b/src/ppx_deriving_runtime.mli
@@ -65,9 +65,9 @@ include (module type of Pervasives with
 module Char : (module type of Char)
 module String : (module type of String)
 module Printexc : (module type of Printexc with
-  type raw_backtrace := Printexc.raw_backtrace and
-  type backtrace_slot := Printexc.backtrace_slot and
-  type location := Printexc.location)
+  type raw_backtrace = Printexc.raw_backtrace and
+  type backtrace_slot = Printexc.backtrace_slot and
+  type location = Printexc.location)
 module Array : (module type of Array)
 module List : (module type of List)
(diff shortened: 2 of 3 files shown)


whitequark:
CI fails.

freebroccolo:
I think the CI is failing because the `Caml` module is only in scope if you are already using `base` and build with `-open Base`. So this isn't a general solution.

You could instead disable the warning per module with `[@@@warning "-3"]` at the top of the module or maybe `[@@warning "-3"]` after the deriving part (not sure the latter will work). Or you could just disable the warning in the build options.

stevejb71:
Fair enough, I'll do that instead. Thanks for the quick response!

whitequark:
It's probably a better idea to bring Format into Ppx_deriving_runtime.

gasche:
We already have `Format` in the runtime, see [src/ppx_deriving_runtime.mli#L103](https://github.com/ocaml-ppx/ppx_deriving/blob/3ae8147/src/ppx_deriving_runtime.mli#L103). So I think that just replacing `Format` with `Ppx_deriving_runtime.Format` in the generated code should work.

stevejb71:
After replacing every Caml with Ppx_deriving_runtime, I get:

```Error: Unbound type constructor Ppx_deriving_runtime.Format.formatter```

I'll look further into it - I'm not very experienced in OCaml so will have to search around.

gasche:
@stevejb71 I looked at the issue, and I think it comes from a mistake in the way `Format` was re-exported in Ppx_deriving_runtime. (It used type substitution `with type formatter := Format.formatter`, which removes the type name from the module, instead of equalities `with type formatter = Format.formatter`.) I just pushed to your PR branch to fix that, and the "show" plugin now builds and passes the tests -- see the updated code.

Could you check that, with this modified PR, there is no warning under your setup?

gasche:
@whitequark, you introduced the type substitutions in b69a830bfc9b3715548fd29e022703aa3273e9bb. Can you confirm that the use of type equalities instead of destructive substitutions sounds correct to you? (See above, the issue with substitutions is that the name is not re-exported in the exported module, so references to the types are incorrect.) Should we consistently use equalities instead of substitutions everywhere?

stevejb71:
I don't see warnings in my build now.

whitequark:
> Can you confirm that the use of type equalities instead of destructive substitutions sounds correct to you?

I think so. I haven't realized `:=` would remove the original `type` instead of turning it into an alias.

gasche:
I rebased the PR on top of a change that consistently uses `with type t = foo` in ppx_deriving_runtime.mli. @whitequark, do you think this is good to merge?

gasche:
Ci is good, so merging. Thanks @stevejb71, @freebroccolo and @whitequark!

PR #170, opened 2018-05-19
Port to Dune

rgrinberg:
This is my WIP pr to port ppx_deriving to dune.

This is not yet done as I need to port `ppx_deriving.main`. But I'm not certain that we actually need it. The users should be able to just list the individual plugins in `-package` and things should just work.

Also, I'm wondering if we really need the dynlinking mechanism for plugins. Most modern users will build a driver out of their plugins. Seems like there should be a mechanism of registering the derivers statically in this case.


whitequark:
> Also, I'm wondering if we really need the dynlinking mechanism for plugins. Most modern users will build a driver out of their plugins. Seems like there should be a mechanism of registering the derivers statically in this case.

Well, if you just remove it, I believe it'll break the pure ocamlfind use case.

rgrinberg:
whitequark <notifications@github.com> writes:

> Well, if you just remove it, I believe it'll break the pure ocamlfind use case.

Not necessarily. At least I think I have an idea on how to address this.
When used with findlib, a deriving plugin can simply be a standalone
preprocessor. When used in conjunction with driver, we can rely on the
registration mechanism. Though it wouldn't have to rely on dynlink
anymore.

doc/style.css
```diff
@@ -1,109 +0,0 @@
-/* A style for ocamldoc. Daniel C. Buenzli */
```
whitequark: Why?

rgrinberg: Well, we're going to be generating the docs with odoc so I don't think this style will work anymore. The default odoc style seems to be inspired by dbuenzli's style for ocamldoc, if that's a small consolation.

whitequark: Alright.

hongchangwu:
It would be great if this can be merged. `ppx_deriving` is a widely used package and a  great addition to the duniverse.

rgrinberg:
This branch is still WIP. There's still some issues I need to fix.

rgrinberg:
cc @emillon

emillon:
LGTM. Is there a particular reason to split the test suites?

rgrinberg:
It makes running the test suite faster. Actually, I should also use the test stanza to simplify some things.

emillon:
Sure! Don't forget to bump the version bound in the opam files.

rgrinberg:
@emillon done

rgrinberg:
Hmm, looks like this still has a problem. Plugins defined using dune cannot be used with the various conv plugins.

rgrinberg:
@diml I'm unable to reproduce this with a small test case. The only way I can reproduce this bug is by pinning this pr and running `$ opam install coq-serapi`. This takes ages unfortunately, but is the only way I know how to repro this.

diml:
I had a look. They were a couple of lines missing in the META file and the `ppx_deriving` binary wasn't installed. I pushed a few commits to update this. I also upgraded the opam file to 2.0.

rgrinberg:
@whitequark This is now ready.

rgrinberg:
Gentle ping to the maintainers @gasche and @whitequark

diml:
BTW, I gave a try at porting ppx_tools to dune and making `ppx_tools.metaquot` work with dune out of the box: https://github.com/ocaml-ppx/ppx_tools/pull/67

With this, we could use ppx_tools.metaquot directly without going through ppxfind.

gasche:
It looks like the `gh-pages` in the Makefile used to update the documentation online is gone? Is this intentional?

Note: I tried to test the documentation output, and `dune build @doc` gives the following warning:

```
        odoc _doc/_odoc/lib/ppx_deriving.api/ppx_deriving.odoc
File "ppx_deriving.cppo.mli", line 185, characters 39-41:
'{P': bad markup
```

The warning is correct (the documentation is using `!{Ppx_deriving_runtime}` instead of the correct `{!Ppx_deriving_runtime}`, but the file location reported is a bit weird.

rgrinberg:
> It looks like the `gh-pages` in the Makefile used to update the documentation online is gone? Is this intentional?

It's not. Let me try and restore it.

> The warning is correct (the documentation is using `!{Ppx_deriving_runtime}` instead of the correct `{!Ppx_deriving_runtime}`, but the file location reported is a bit weird.

Perhaps, it's due to the preprocessor outputting wrong locations? In any case, not much can be done about this. I will fix the markup though.

gasche:
(I fixed the doc markup issue in master.)

rgrinberg:
Alright. I tried restoring the gh-pages target.

diml:
Shouldn't we let `dune-release` handle the upload of the doc? It's less specific code to maintain here.

rgrinberg:
I'm in favor of however way @gasche and @whitequark would like to maintain the docs.

By the way, once this PR is merged I can prepare the PR that can consolidate this repo with ppx_deriving_protobuf and ppx_deriving_yojson. @emillon @NathanReb that should simplify maintenance of all these derivers for ya'll

gasche:
I looked at this PR again, rebased it on top of master (to get the doc-markup fix that was showing in the new `make doc` target). I'll merge it when the CI returns and is green.

My gut feeling is that there remain horrible problems hiding under the surface that will break some ppx_deriving plugins or their user if they are not also use Dune -- time will tell. That was an issue that gave me pause when the PR was first proposed, but it is much less of an issue now, given that almost everything is using Dune these days. So we have a lot of users whose life will be made moderately nicer (and also us developers with faster build, nicer doc etc.), and a few users whose life might be made worse until they move on to Dune themselves. It's a deal.

rgrinberg:
Thanks @gasche. This is one of those PR's where the consequences are hard to foresee. I plan on helping to support this if any dune related bugs are filed. I'm confident that this will not be nearly as bad as when omp/dune initially came out.

gasche:
Merged. Woo! Thanks @rgrinberg (and @diml) for your work. I have to do a release now...

gasche:
Just as a cross-comment, there is an enlightening discussion on the difference between `ppx_deriving` and  `ppx_deriving.runtime` and `ppx_deriving.api` before and after the Dune port in [the pre-release opam-repository PR#13652](https://github.com/ocaml/opam-repository/pull/13652). The summary is that we may have to change our plugin's opam metadata (and possibly, in some rarer cases, fix the plugin build scripts) to fix dependency issues that were hidden previously, and now become visible as Dune is stricter about not mixing `.cmi` of different libraries together.

gasche:
I'll summarize here my complete understanding of the issue, because I think it is better to keep the content in here rather than opam-repository.

There are three pieces to `ppx_deriving`, both before and after the Dune port (no change)
- `ppx_deriving` declares stuff to ocamlfind (ocamlfind users need a (possibly transitive) dependency on `ppx_deriving` for the ppx_deriving plugins they depend on to be linked correctly by ocamlfind)
- `ppx_deriving.api` contains the `Ppx_deriving` module, which plugin authors depend on to implement their plugins
- `ppx_deriving.runtime` contains the `Ppx_deriving_runtime` module, which plugin *users* depend on to compile the plugin-generated code

In an ideal world, users should never need to explicitly indicate a dependency on any of these things, just on the plugin(s) that they use. Plugin authors should depend on `ppx_deriving.api` in their build scripts, and export in their metadata a dependency to `ppx_deriving` (for the ocamlfind support) and `ppx_deriving_runtime` (for everyone).

(Note: If the plugin authors use dune, the dependency on the runtime should go through a `ppx_runtime_libraries` field in the `library` stanza for their plugin; if they have a runtime module of their own that depends on `ppx_deriving.runtime`, indicating their own runtime module suffices. If they don't, they have to ensure that the `ppx_deriving.runtime` dependency is properly documented in their META  files. In both case, the dependency should be in the opam file.)

With the pre-dune system, the `Ppx_deriving` and `Ppx_deriving_runtime` would put their `cmi` files at the same place (exactly at `$(ocamlfind query ppx_deriving)`); with Dune, they get installed in separate sub-directories `api` and `runtime`. As a consequence, two kind of metadata mistakes could go unnoticed before:

- If a user would specify a dependency on `ppx_deriving` in their build scripts, they would get the runtime for free, so the plugin authors could forgot to express a dependency to `ppx_deriving_runtime` and things would silently work. We can fix this by adding the proper dependencies in plugins. (In general changing the opam metadata should suffice, unless they use dune but forgot to specify they `ppx_runtime_libraries` correctly.)
- if a plugin author would specify a dependency on `ppx_deriving` in their build script, they would have access to `Ppx_deriving` module for free, and not notice that a dependency on `ppx_deriving.api` is missing. We can fix this by adding the proper dependencies in plugins, but this may require a change to their build scripts (in addition to opam metadata).

Unrelatedly to all this, we got rid of a `cppo_ocamlbuild` dependency that some of our users (in particular plugins) would get transitively, and thus forget to specify it themselves. We can fix this by fixing their opam metadata.

diml:
Yes, that seems right. BTW, if we want a bit more time to deal with upgrading packages, we can take the same approach lwt took: we write a post-build script that rewrites the .install and META files to get rid of the `api` and `runtime` sub-directories. Then we communicate about the upcoming breaking change and we only get rid of this script in ppx_deriving 5.0.

gasche:
I'm tempted to deal with the packages that can be fixed by opam-repository-level changes right now before the release, and see what else breaks. If it's not too much, I'll add an upper-bound on ppx_deriving for these packages and release.

Issue #175, opened 2018-12-14
Missing plugin(native) entry in META's runtime package?


vprevosto:
I've been trying to dynamically load a `.cmxs` that depends on `ppx_deriving_yojson`, thus on `ppx_deriving`, using `findlib` to get all necessary dependencies loaded before my actual file. This failed when `findlib` said that `ppx_deriving_runtime.cmxa` should be loaded, instead of `ppx_deriving_runtime.cmxs`. Adding the line

```
plugin(native) = "ppx_deriving_runtime.cmxs"
```

as part of the `runtime` package description in `ppx_deriving`'s `META` file solved the issue.

Did I overlook something, or is this really an issue in `META`?

## steps to reproduce: 

Compile the following code with `ocamlfind ocamlopt -package findlib -package dynlink -linkpkg -linkall -o load load.ml` and launch `./load` (shortened from the mechanism used by Frama-C to load its plugins).

```ocaml
let () = Findlib.init ()

let () = Dynlink.init ()

let () = Dynlink.allow_unsafe_modules true

let predicates =
  ["plugin", ["native"]; "archive", ["plugin"]; "archive", ["native";"plugin"]]

let package = "ppx_deriving.runtime"

let find_file package acc (var, predicates) =
  match acc with
  | None ->
    (try Some (Findlib.package_property predicates package var)
     with Not_found -> None)
  | Some _ -> acc

let load package =
  let file = List.fold_left (find_file package) None predicates in
  match file with
  | Some file ->
    let base = Findlib.package_directory package in
    let full_path = Findlib.resolve_path ~base file in
    (try Dynlink.loadfile full_path
     with Dynlink.Error e ->
       Printf.eprintf "Dynlink error while loading %s: %s\n%!"
         full_path (Dynlink.error_message e))
  | None -> Printf.eprintf "No file found for loading %s\n%!" package

let () = List.iter load (Findlib.package_deep_ancestors ["native"] [package])
```

whitequark:
I *very* vaguely recall some issue with native plugins that prevented this from working. But if it actually works, it's probably just a missing entry.

vprevosto:
Thanks for your answer.

[...]

PR #176, opened 2018-12-19
[META] add plugin(native) directive for runtime

vprevosto:
One-line fix in `META` for #175

### DIFF
--- a/pkg/META.in
+++ b/pkg/META.in
@@ -9,6 +9,7 @@ package "runtime" (
   description = "Runtime component of built-in derivers"
   archive(byte) = "ppx_deriving_runtime.cma"
   archive(native) = "ppx_deriving_runtime.cmxa"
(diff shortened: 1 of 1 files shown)


gasche:
I apologize for being very slow on that PR, but my understanding is that the port to Dune that was just merged should magically fix that, by generating the right META files for us. Would you by chance be able to test that assumption?

(Some of the existing plugin's opam metadata may have to be changed slightly to accommodate some stricter rules about not-mixing-cmi-files from Dune, so it might be that your test fails because of a missing `ppx_deriving.api` or `ppx_deriving.runtime` dependency in the plugin you test.)

vprevosto:
Thanks. I have tested the plug-in against the master branch of `ppx_deriving` and everything looks fine. The `META` file of `ppx_deriving` seems indeed to include the necessary entry for the runtime. I guess that this PR can thus be closed.

gasche:
Great! Closing.

Issue #171, opened 2018-06-25
Unused values when deriving show for private types


emillon:
Hi,

The following piece of code triggers warning 32 (unused value `show_x`). This is a common pattern, it happens whenever an exposed type uses a type that is not exposed.

```ocaml
module M : sig
  type t
  [@@deriving show]

  val make : int -> t
end = struct
  type x = int
  [@@deriving show]

  type t = C of x
  [@@deriving show]

  let make n = C n
end
```

The problem is the following:
- 4 bindings are created: `pp_x`, `show_x`, `pp`, and `show`.
- `pp`, and `show` are exposed
- `pp_x` is used in the generated definition for `pp`
- `show_x` is not, so a warning is triggered.

The root cause is similar to ocaml-ppx/ppx_deriving_yojson#76: there is a "convenience function" (here `show`) that is not meant to be used recursively.

I am not sure how to best fix this without breaking existing code, but here are some possibilities:
- generate `let _ = show_x` to suppress the warning
- similarly, wrap the generated structure within `[@@@warning "-32"]`
- add an option so that we can `[@@derive show { only_pp = true }]`
- introduce a `[@@deriving pp]` that only derives `pp`. This would enable `[%pp: t]` as well.

What do you think?

Thanks!

whitequark:
I think `[@@warning "-32"]` and `[@deriving pp]` both seem like good options; we can simply do both of them.

NathanReb:
Hi,

What's the status on this issue?

I'm happy to contribute for the `[@deriving pp]` if that's fine by you!

whitequark:
@NathanReb As mentioned above in https://github.com/ocaml-ppx/ppx_deriving/issues/171#issuecomment-400598844, I think both `[@@warning "-32"]` and `[@deriving pp]` ought to be implemented. If you can do one of them, great!

PR #179, opened 2019-02-07
Suppress warning 32 for show

NathanReb:
As reported in #171, using `[@@deriving show]` can trigger unwanted unused value warnings.

This PR adds a `[@@warning "-32"]` attribute to the generated `let show = ...` structure item to prevent such warnings.

I might look into a `[@@deriving pp]` alternate solution if I get some time to do so but this seemed like a simpler and faster first step.

This is my first PR on `ppx_deriving` so I might have gotten a few things wrong, especially around the `loc` to use for the generated code.

I tested this on our codebase at Cryptosense and it seems to do the job!

I guess it might be a good idea to expose a way to disable this feature since that makes it impossible for `pp` and `show` to trigger such a warning even when it would be relevant.

Let me know what you think!


NathanReb:
Just rebased on top of #180 to see if it would pass the CI.

NathanReb:
Rebased on top of #170 as it does fix the CI.

PR #182, opened 2019-03-14
Split all stanzas into separate dirs

rgrinberg:
This makes the generated .merlin completely accurate and also eases
understanding the project structure


gasche:
Pre-dune had a `ppx_deriving.std` package to aggregate dependencies on all built-in derivers at once. Does this still exist in the dune system? (I would expect it to be declared in `src_plugins/dune` in/after your PR, and if I read the diffs correctly this file is gone.)

rgrinberg:
It's still there. I put it in `src_plugins/std`, since it is named `ppx_deriving.std` - on the same level as the other plugins.

gasche:
Merged, thanks!

PR #188, opened 2019-05-13
4.02 nit in the tests

ELLIOTTCABLE:
A couple ppx_deriving.map tests used `String.uppercase_ascii`. This was the only thing preventing the library from building and passing tests on 4.02.3; and it looked like, at their callsites, reliable behaviour w.r.t. higher-plane bytes wasn't important to the semantics of the test.

Let me know if there's a better solution!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ _build
 pkg/META
 src_test/_tags
 .merlin
+_opam
--- a/src_test/map/test_deriving_map.cppo.ml
+++ b/src_test/map/test_deriving_map.cppo.ml
@@ -100,7 +100,7 @@ let test_var2 ctxt =
   assert_equal ~printer:(show_var2 fmt_int) (A2 5) (map_var2 int_of_float (A2 5.))
 
 let test_var3 ctxt =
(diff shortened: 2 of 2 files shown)


rgrinberg:
This solution isn't good because it will introduce deprecation warnings. If you want to use these functions, you should add the appropriate attributes to silence the warnings.

ELLIOTTCABLE:
@rgrinberg I'm not familar — glancing at the manual, and I'm actually having some trouble googling up information about how to do this. I'm assuming it's a specially-formatted comment before the offending line?

(And yeah, I just double-checked on 4.06.1, and this totally adds a bunch of deprecation warnings. Yikes! Good catch!)

rgrinberg:
I tested it and it works. However, I'm quite surprised that this deriver even works on 4.02. I thought ppx_deriving and the driver stack moved to a later OCaml long ago.

.gitignore
```diff
@@ -5,4 +5,5 @@ _build
 *.install
 pkg/META
 src_test/_tags
-.merlin
\ No newline at end of file
+.merlin
+_opam
```
gasche: This change is unrelated but still good to have, thanks.

PR #193, opened 2019-06-25
4.08 support

gasche:


XVilka:
Sorry, obviously https://github.com/ocaml-ppx/ppx_deriving/pull/195 should be a part of this.

gasche:
I included #195 in the PR, thanks! Let's see what the CI results are.

src/runtime/ppx_deriving_runtime.cppo.ml
```diff
@@ -0,0 +1,65 @@
...
+type nonrec 'a array = 'a array
+type nonrec 'a list = 'a list
+type nonrec 'a option = 'a option
+type nonrec nativeint = nativeint
+type nonrec int32 = int32
+type nonrec int64 = int64
+type nonrec 'a lazy_t = 'a lazy_t
+type nonrec bytes = bytes
+
+#if OCAML_VERSION >= (4, 08, 0)
```
ELLIOTTCABLE: So, this is merged and published, so I don't know if this is the best place to raise this — but using `cppo` in the *runtime*, specifically, is going to cause me … a whole, whole lot of problems over in `bs-deriving` land (not in the least, because `#if OCAML_VERSION` actually [*already means something*](https://bucklescript.github.io/bucklescript/Manual.html#_conditional_compilation_support_static_if) over there — with a different semantic!). )=

Not y'all's problem, of course; but before I go try and make my rickety, arcane build-system about ten times as rickety and arcane as it already is … is there any chance there's an easy, alternative way to do this upstream? Since this is the only occurrence in the runtime?

gasche: I'm happy to try to make the bs-deriving life easier, but I'm worried that there may not be a very easy way to avoid having two different logics here -- it's solving several delicate problems (deprecation of `Pervasives`, semantics of `include`, etc.) and it is actually scoping over most of the runtime definition, not just a small part of it.

What would be your recommendation for a conditional-compilation approach that works with both cppo and bspp? Would you avoid all in-file logic and use two different `.ml`, selected by the build system?

Maybe it is possible to use both build systems to define one variable (`OCAML_VERSION_GTE_408`), and use a test on that variable in a syntax that is compatible with both preprocessors?

ELLIOTTCABLE: Hmmmmmmmmmm. Well, I'd like to avoid "partially upstreaming" bs-deriving (in the sense of making weird changes like adding `OCAML_VERSION_but_not_because_some_other_thing_not_related_to_this_codebase_is_weird`, that you've then gotta document somewhere, test, avoid breaking in the future …)

If `OCAML_VERSION` is idiomatic in the way it's used here, I'm not sure what to do; I'm not exactly a high-wizard of OCaml, unfortunately, just a dude gluing some stuff together. Poorly. 🤣

Two files might work, but is still teetering on just … upstreaming the maintenance burden. I was mostly, vaguely hoping that “hey! wow! there just happen to be two equally-good ways to do this upstream, and the other one won't be as big a problem for Elliott! Yay!”, but doesn't look like it.

I'm gonna keep this in mind and ruminate on it for a while; maybe I can figure something simple out in the coming weeks. (Thanks for the quick replies, by the way!)

gasche: I don't personally mind having some changes in the upstream repo that are solely justified by bs-deriving, especially if they only touch the implementation (so we have to document the code but the end-users will not see a difference). I would keep the extra-maintenance-burden in mind, but maybe there is a way that works and isn't actually that burdensome.

(Yes, `OCAML_VERSION` is used in the idiomatic way here. I think it would have made sense for the Bucklescript people to use the de-facto standard syntax of cppo in their preprocessor syntax, but well...)

ELLIOTTCABLE: (Note that I've opened BuckleScript/bucklescript#3820 about this.)

PR #195, opened 2019-06-26
Add autobuilds for OCaml 4.08

XVilka:


gasche:
This is now included in #193 and both should be merged at the same time -- so I'm closing here. Thanks for the patch!

gasche:
For reference, the 4.08 build fails on Travis because the updated ppx_tools branch is not yet part of a release. (Which is probably just as good given that I found a small bug/omission in it.)

PR #197, opened 2019-07-06
Add `Result.result` to the .mli file

anmonteiro:
The comment in the `.ml` file mentions this is done for backwards compatibility but it wasn't actually exposed in the interface.


gasche:
Thanks! That's indeed an omission on my part. I'm merging this right now, so you should be able to rebase https://github.com/ocaml-ppx/ppx_deriving_yojson/pull/99 to remove the conditionals on that part of the PR.

PR #200, opened 2019-07-08
Result: ensure Result.t and Result.result are exposed under both 4.08 and earlier versions

gasche:
cc #197 @anmonteiro . Would that change avoid any Result-related code change in 
https://github.com/ocaml-ppx/ppx_deriving_yojson/pull/99 ?

My goal is not only to fix ppx_deriving_yojson or make our life easier, but to ensure that we limit breaking other plugins as much as possible. For that reason, even the partial workaround that you used in https://github.com/ocaml-ppx/ppx_deriving_yojson/pull/99/commits/3e16cd22d6c410c7d5362ed68334edda742094e9 is not fully satisfying.

### DIFF
--- a/src/runtime/ppx_deriving_runtime.cppo.ml
+++ b/src/runtime/ppx_deriving_runtime.cppo.ml
@@ -23,6 +23,16 @@ type nonrec bytes = bytes
 module Stdlib = Stdlib
 
 include Stdlib
+
+module Result = struct
+  type ('a, 'b) t = ('a, 'b) Result.t =
+    | Ok of 'a
+    | Error of 'b
+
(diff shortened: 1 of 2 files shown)


gasche:
(I would like to find a way to resolve this general `Result` issue, and once this is done I'll cut out a 4.08 release.)

gasche:
I just checked that, from this branch, https://github.com/ocaml-ppx/ppx_deriving_yojson/commit/cbc50080999f6c51b1030e8ae0939b2c14cec814 (the tip of the 4.08 changes *without* the Result-related changes) is building fine. Merging.

PR #201, opened 2019-07-08
release 4.4 with 4.08 support

gasche:


gasche:
See https://github.com/ocaml/opam-repository/pull/14454 for the pre-release opam CI testing -- everything passes, but "everything" is not much given that most dependents are still blocked on 4.08.

PR #206, opened 2019-09-09
Migrate to ppxlib

thierry-martinez:
- Lower bound on OCaml version is raised to 4.04.1 because ppxlib
  0.9.0 requires it.

- ppx_tools is no longer used, because it is not compatible with
  ppxlib. ppx_tools was mostly used (1) for metaquot, which is handled
  by ppxlib.metaquot, and (2) for the Ast_convenience module, which is
  now defined in Ppx_deriving module.

- Most cppo conditional directives are gone, either because they
  were there to support OCaml <4.04.1, or because they are subsumed
  by the Migrate_parsetree.OCaml_408 implied by opening Ppxlib.


gasche:
Ẁow, thanks a lot! I'm looking forward to having a look at this -- I might need a few more days before I have time for a proper review.

I don't have much experience working with ppxlib project. One thing that I find a bit surprising is that, for most files, you never explicitly state that you want the 4.08 version of the AST (some other modules are explicitly accessed with a `Ppxlib.Foo.408.Bar` path). I would have expected an explicit selection of the 4.08 version of the AST, to be robust to future changes of "default AST version" in ppxlib. Is it the recognized best practice to keep the AST versioning mostly-implicit? I guess that means we have to release new versions if the ppxlib default changes? (Should we also have an upper-bound on the supported ppxlib version then?)

rgrinberg:
I can reproduce the stackoverflow in CI:

```
         ppx src_test/create/test_deriving_create.pp.ml (exit 2)
(cd _build/default && .ppx/ec7ca9c12b6a84c8df3dae3b4f776e51/ppx.exe -o src_test/create/test_deriving_create.pp.ml --impl src_test/create/test_deriving_create.ml -corrected-suffix .ppx-corrected -diff-cmd - -dump-ast)
Fatal error: exception Stack overflow
Raised at file "src/ast_408_helper.ml", line 14, characters 40-47
Called from file "ppx_deriving.cppo.ml", line 749, characters 8-173
Called from file "src/ast_408.ml", line 3132, characters 51-73
Called from file "src/ast_408.ml", line 3199, characters 10-35
Called from file "src/ast_408.ml", line 3495, characters 41-73
Called from file "ppx_deriving.cppo.ml", line 672, characters 17-21
Called from file "ppx_deriving.cppo.ml", line 728, characters 44-144
Called from file "ppx_deriving.cppo.ml", line 741, characters 20-59
Called from file "src/migrate_parsetree_driver.ml", line 249, characters 21-60
Called from file "src/migrate_parsetree_driver.ml", line 255, characters 4-82
Called from file "src/driver.ml", line 873, characters 53-82
Called from file "src/driver.ml", line 1269, characters 6-24
Re-raised at file "parsing/location.ml", line 904, characters 14-25
Called from file "src/driver.ml", line 1272, characters 4-59
Called from file ".ppx/ec7ca9c12b6a84c8df3dae3b4f776e51/_ppx.ml", line 1, characters 9-36
         ppx src_test/deriving/test_ppx_deriving.pp.ml (exit 2)
```

src/api/ppx_deriving.cppo.ml
```diff
@@ -17,28 +8,122 @@
...
-#define Rinherit_patt(typ) {prf_desc = Rinherit(typ); _}
-#endif
-
-open Longident
 open Location
 open Asttypes
-open Parsetree
 open Ast_helper
+
+module Ast_convenience = struct
```
rgrinberg: Do we really need to redefine this? As far as I see, we are upgrading the version of the AST used by ppx_deriving - that is already a breaking change. We might as well ask users to switch to the ppxlib API in their plugins as well.

thierry-martinez:
> OCAML_VERSION >= (4, 03, 0)

Thanks, such comparisons are now removed systematically (commit 5d050c).

thierry-martinez:
> I can reproduce the stackoverflow in CI:
> 
> ```
>          ppx src_test/create/test_deriving_create.pp.ml (exit 2)

Oops. My new code based on Ppxlib.map was incorrect (even if I was not able to figure why exactly), I restored the previous code based on Ast_mapper (even if it is now somehow discouraged by ppxlib).

src/api/ppx_deriving.cppo.ml
thierry-martinez: The idea was mainly to reduce the amount of changes elsewhere in the library (because `Ast_convenience` is used a lot of times in standard plugins).

thierry-martinez:
(Sorry for the very late answer: holidays, etc.)

> One thing that I find a bit surprising is that, for most files, you never explicitly state that you want the 4.08 version of the AST (some other modules are explicitly accessed with a `Ppxlib.Foo.408.Bar` path).

As far as I understand, each `ppxlib` is tied with one specific OCaml AST version. I think that the only explicit references to 4.08 is to access `Migrate_parsetree` modules, especially to have access to `Ast_mapper` that is shadowed by `ppxlib`.

> Is it the recognized best practice to keep the AST versioning mostly-implicit?

I think yes.

> I guess that means we have to release new versions if the ppxlib default changes? (Should we also have an upper-bound on the supported ppxlib version then?)

The previous times that ppxlib bumped OCaml AST versions, upper-bounds were added to all the packages that used the previous version in opam-repository. I think that it is better to wait for the next AST bump before being explicit about the upper-bound, otherwise the package will become uninstallable each time ppxlib is updated (even if the AST version remains stable).

gasche:
So, if I understand correctly, ppxlib fixes a given "recent enough" compiler version that it supports well, and deals with all other versions (older or more recent) through o-m-p conversion. Migrating to this model has a downside: users of OCaml versions more recent than ppxlib's standardized version will not be able to use their new syntactic constructs (in types) before ppxlib globally upgrades its version.

Before, on each new OCaml release, ppx_deriving would have to get work to update to the new AST format (conditionally), and then users could use it (including new features) as soon as that work was done. After the switch, we don't have to do much on new OCaml versions, but users have to wait for ppxlib to upgrade to use recent features (for mere refactoring of the AST, o-m-p has their back). This upgrade is done infrequently as it breaks all ppxlib users.

On the other hand, in practice before I would ensure that ppx_deriving *compiles* with the new OCaml version, but not that it actually supports new AST features (and: radical new changes to the type-declaration AST are fairly rare anyway). So if you tried to use one of those shiny recent features, chances are you would get into an `assert false` at preprocessing time.

XVilka:
Well, seems that ppxlib is just a temporary solution anyway, before [new ppx](https://github.com/ocaml-ppx/ppx) is ready.

gasche:
But we don't know when the new ppx will be ready, and ppxlib already brings benefits today, so adopting ppxlib now makes a lot of sense.

XVilka:
@gasche yes. I just meant that it will break older versions anyway, once migration to ppx is finished.

src/api/ppx_deriving.cppo.ml
```diff
@@ -507,11 +561,13 @@ let fold_exprs ?unit fn exprs =
     | None -> raise (Invalid_argument "Ppx_deriving.fold_exprs")
 
 let seq_reduce ?sep a b =
+  let loc = !Ast_helper.default_loc in
```
gasche: I wondered whether we can give a better location to this (join the locations of `a` and `b`). In fact `default_loc` is already a reasonable choice, as the ppx_deriving mapper already calls `Ast_helper.with_default_loc` to set the default location when generating code from a declaration -- the location points to the whole declaration that is currently being processed. Good enough.

src/api/ppx_deriving.cppo.ml
```diff
@@ -762,5 +807,5 @@ let hash_variant s =
    driver. *)
 let () =
   Migrate_parsetree.Driver.register ~name:"ppx_deriving"
-    (module Migrate_parsetree.OCaml_current)
+    (module Migrate_parsetree.OCaml_408)
     (fun _ _ -> mapper)
```
gasche: Note: `api/ppx_deriving` here registers the mapper into the o-m-p driver (for use with other o-m-p-enabled transformations). In `src/ppx_deriving_main` there is another registration mechanism, using `Ast_mapper` from the compilerlibs, which is only used when producing an executable-style preprocessor (instead of being linked as a plugin of a driver executable). The two registration mechanisms seem to cohabit nicely enough, and in any case were already present before this PR.

src_plugins/create/ppx_deriving_create.cppo.ml
```diff
@@ -1,11 +1,9 @@
-#include "../compat_macros.cppo"
-
-open Longident
+open Ppxlib
 open Location
 open Asttypes
 open Parsetree
 open Ast_helper
-open Ast_convenience
+open Ppx_deriving.Ast_convenience
```
gasche: We now have to use our own redefinition of `Ast_convenience`, so plugin code has to be adapted slightly. (Note: *not* redefining `Ast_convenience` would mean cumbersome changes for third-party plugin authors.)

src_plugins/create/ppx_deriving_create.cppo.ml
```diff
@@ -1,11 +1,9 @@
-#include "../compat_macros.cppo"
-
-open Longident
```
gasche: `Longident` is not opened anymore. @thierry-martinez explains that we could not open Ppxlib.Longident in the global scope because that one shadows `(=)` from the standard library (to be monomorphic on longidents), which we use in plugin code.

src_plugins/create/ppx_deriving_create.cppo.ml
gasche: The most notable change for this plugin is that essentially no (other) code change is required for plugin authors. (The patchset only get rids of now-useless IFDEFs.)

src_plugins/enum/dune
```diff
@@ -9,6 +8,6 @@
  (public_name ppx_deriving.enum)
  (synopsis "[@@deriving enum]")
  (preprocess
-  (action (run ppxfind -legacy ppx_tools.metaquot --as-pp %{input-file})))
- (libraries compiler-libs.common ppx_tools ppx_deriving.api)
+  (pps ppxlib.metaquot))
+ (libraries compiler-libs.common ppxlib ppx_deriving.api)
```
gasche: Plugin authors will have to change their build system in a similar way to this. Releases of the plugin with this new style will not be compatible with older versions of ppx_deriving -- but they are compatible with older OCaml releases. (@thierry-martinez explains: the reason why we need a change is that ppxlib's metaquot is fixed on 4.08, while ppx_tools.metaquot works on the current OCaml version.)

gasche: So plugin authors will have to add an upper bound on their older releases to the latest non-ppxlib version, and a lower bound on their new ppxlib-friendly releases on the new ppx_deriving version. But both kind of versions will support existing OCaml releases, and the code changes are very small in simple cases.

src_plugins/enum/ppx_deriving_enum.cppo.ml
```diff
@@ -77,7 +63,9 @@ let mappings_of_type type_decl =
     match mappings with
     | (a, { txt=atxt; loc=aloc }) :: (b, { txt=btxt; loc=bloc }) :: _ when a = b ->
       let sigil = match kind with `Regular -> "" | `Polymorphic -> "`" in
-      let sub = [Location.errorf ~loc:bloc "Same as for %s%s" sigil btxt] in
+      let sub =
+        [Ocaml_common.Location.errorf
+          ~loc:bloc "Same as for %s%s" sigil btxt] in
```
gasche: Note: we will have to document this small change patterns to third-party plugin authors.

rgrinberg:
> Migrating to this model has a downside: users of OCaml versions more recent than ppxlib's standardized version will not be able to use their new syntactic constructs (in types) before ppxlib globally upgrades its version.

Doesn't ppx_tools already force ppx_deriving's hand this way?

> On the other hand, in practice before I would ensure that ppx_deriving compiles with the new OCaml version, but not that it actually supports new AST features (and: radical new changes to the type-declaration AST are fairly rare anyway). So if you tried to use one of those shiny recent features, chances are you would get into an assert false at preprocessing time.

In practice, Users of ppx_deriving should not be recommended to upgrade to a version of OCaml until ppxlib worked well on it.

In my opinion, we should switch to ppxlib. In ppx, there's going to be features that should make the transition easier and ppxlib will not be broken overnight once ppx is released.

gasche:
> Doesn't ppx_tools already force ppx_deriving's hand this way?

The difference, if I understand correctly, is that as soon as ppx_tools is released for a new OCaml version, all constructs of this new version work correctly with the ppx_tools codebase. For ppxlib, as long as ppxlib's own AST versoin is fixed to 4.08, releases that provide newer-version support will let users use ppx_deriving, but not in combination with the not-in-4.08 features. ppxlib could upgrade to a more recent OCaml version, but this is (if I understand correctly) done somewhat infrequently, to limit the amount of code churn (... otherwise ppxlib users would have to adapt their code at each OCaml release, which is precisely what they don't want).

rgrinberg:
Okay, then I think we're on the same page. Just to confirm: I don't think it's in the interest of (most) users to follow changes in the compiler as soon as possible. A more controlled upgrade process as in ppxlib would be more preferable for them.

src/api/ppx_deriving.cppo.mli
```diff
@@ -1,12 +1,9 @@
 (** Public API of [ppx_deriving] executable. *)
 
-open Parsetree
+open Ppxlib
```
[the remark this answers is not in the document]
rgrinberg: > Furthermore, ppxlib provides its own deriving mechanism; it is incompatible with current ppx_deriving plugins, but we could consider turning ppx_deriving into a compatibility layer (for plugins) on top of ppxlib-deriving.

Indeed. In fact, I've added some features in ppxlib to make this exactly possible. For example, ppxlib now supports the same quoting mechanism as in ppx_deriving, and deriving on module types.

src/api/ppx_deriving.cppo.ml
rgrinberg: Would it be possible to use ppxlib's registration mechanism? Even if you don't use the context free api, it still has some advantages.

gasche:
If I'm a user of ppx_deriving and its plugin, what I want is for this code to work on my codebase whatever I do. (But of course this may require infinite amount of work for the maintainers of ppx_deriving and its plugins.) It's never in the interest of those downstream users to be told "nope, this feature is not supported yet".

If I'm a plugin author, or maintainer of ppx_deriving (both are in the same boat), I have to make a choice between (1) doing the upgrade to newer OCaml features ourselves, which makes our user maximally happy (but requires maximal work), or (2) relying on a compatibility layer such as o-m-p and ppxlib, which moves the decision to other people. With o-m-p alone, I can expect good compatibility with new features a couple weeks after each new release (so my users don't really see the difference with me putting the work). With ppxlib, I tell them that I'm bound on 4.08, and I won't change except every few years; convenient for me, but choices are taken away from them (and also from me).

src_plugins/enum/dune
rgrinberg: Is `compiler-libs.common` still needed as a dependency here btw? Shouldn't we be using `ocaml-compiler-libs`?

rgrinberg:
> doing the upgrade to newer OCaml features ourselves, which makes our user maximally happy (but requires maximal work)

This makes users happy in the *short term*, especially if they're just kicking the tires off a new release. Most users are much better off waiting until ppxlib (and similarly merlin, ocamlformat, dune, etc.) are updated to support the features they want. Yes, they might have their concrete problems with ppx_deriving solved earlier, but it's quite likely they'll run into other related problems.

Anyways, I think we're in violent agreement regarding this PR regardless of the details :)

gasche:
I agree that we are in agreement, but I remain confused by your point.

Merlin, ocamlformat, dune already support the full 4.09 language, and will support the full 4.10 shortly. My point is that ppxlib, by design, will stay longer on an older OCaml release (4.08), offering only partial compatibility with newer language features. I think this has a clear cost that extension authors deciding to rely on ppxlib need to be aware of, and it cannot be turned into a good thing; it's a downside (that comes with other positive aspects of the tools).

rgrinberg:
The latest version of merlin does not support 4.09.

As for the other tools, it's definitely a bit harder to bring examples from the past that aren't parse tree related. Let's take the upcoming `-I` for individual cmi files for example. Obviously dune will eventually support this, but it could be delayed by unforeseen events. In the meantime, I would recommend users to just be patient and wait for this feature. There's certainly a downside to this, but it's hard to imagine a sensible alternative.

I'm not really sure ppxlib stays behind by design by the way. It's just quite a bit larger than omp so it's not as easy to port it.

Anyhow, let's switch gear and discuss this PR :) Do you think we should port the common plugins to work with this PR before making a release? I would think that `ppx_deriving_yojson` should be ported at the very least.

gasche:
Yes, I was planning to port some plugins, and then use the opam-ci to observe the revdep breakage and clear them before the release. (But for some plugins we can just add an upper bound on older versions, without making a compatible release.). It is possible that unforeseen compatibility issues would delay the release.

rgrinberg:
I have a port of ppx_deriving_yojson to use ppxlib and the version of ppx_deriving in this branch. Things are working fine and I'm planning to do a similar port of ppx_deriving_protobuf. I'll put up some PR's shortly.

rgrinberg:
In fact, since this PR completely eliminates the ifdef's we should just remove cppo entirely. We can rename the filenames to be just .ml and .mli, and remove cppo from the opam file.

thierry-martinez:
AST is now converted at the level of each type declaration when `deriving` is used instead of converting the whole AST. Commits are squashed. I think that this PR is now ready for merging.

src/api/ppx_deriving.cppo.ml
```diff
@@ -640,43 +728,48 @@ let mapper =
...
+          copy_deriver deriver typ)
       | _ -> Ast_mapper.(default_mapper.expr) mapper expr
       end
     | _ -> Ast_mapper.(default_mapper.expr) mapper expr
   in
   let structure mapper items =
     match items with
     | { pstr_desc = Pstr_type(_, typ_decls) as pstr_desc ; pstr_loc } :: rest when
-        List.exists (fun ty -> has_attr "deriving" ty.ptype_attributes) typ_decls
+        List.exists (fun ty -> has_attr_current "deriving" ty.ptype_attributes) typ_decls
```
gasche: I think it would make sense to shadow `has_attr` with `has_attr_current` at the beginning of `mapper`, so that these don't have to change.

rgrinberg:
I believe there's still a bit of cppo that can be removed after the switch to ppxlib. For example:

```
#if OCAML_VERSION >= (4, 05, 0)
  tyvar.Location.txt
#else
  tyvar
#endif
```

In `test_api.cppo.ml`.

rgrinberg:
I'm also in favor of removing all these macros now:

```
#define Attribute_expr(loc_, txt_, payload) { attr_name = \
                                                { txt = txt_; loc = loc_ }; \
                                              attr_payload = payload; \
                                              attr_loc = loc_ }
#define Attribute_patt(loc_, txt_, payload) { attr_name = \
                                               { txt = txt_; loc = loc_ }; \
                                              attr_payload = payload; \
                                              attr_loc = _ }
```

rgrinberg:
Also, we should remove the unsupported versions of OCaml from CI.

XVilka:
Would be nice to retest with the [newer ppxlib-0.10.0](https://github.com/ocaml-ppx/ppxlib/releases/tag/0.10.0) released recently.

XVilka:
Any updates on this one?

rgrinberg:
@XVilka if you are dependent on this work, would you mind helping out? We could definitely use some help here to get it out faster.

XVilka:
@rgrinberg ok, will do

gasche:
For the PR inside ppxlib itself, I think that not much more needs to be done than me taking a last look and clicking the "merge" button. Then we need to port the outside-tree plugins to the new API, and also to be careful in the release process to detect compatibility failures (with plugins or with user code); there will be a ton of revdeps to analyse carefully, and of course help will be more than welcome then.

XVilka:
@gasche @rgrinberg there are some cppo macroses left, to support 4.04. What do you think about dropping 4.04 entirely and do minimum requirement of 4.05?

See the particular example of https://travis-ci.com/radareorg/radare2/jobs/267557515#L1867

rgrinberg:
I think it's a good idea because the old version of ppx_deriving should still be available to users of 4.04.

Are you referring to the macros that I mentioned? I don't think those are needed to support 4.04. When we use ppxlib, we are using the version of Ast that it gives us, and there's no need to conditionally compile for other versions of the Ast.

XVilka:
@rgrinberg no, there are some leftovers in tests

XVilka:
I created a new pull request since cannot push here https://github.com/ocaml-ppx/ppx_deriving/pull/210

thierry-martinez:
Continued in #210 .

PR #211, opened 2019-12-23
Add support for OCaml 4.10

kit-ty-kate:
Relying on https://github.com/ocaml-ppx/ppx_tools/pull/79, here is a PR to add support for OCaml 4.10


kit-ty-kate:
ppx_tools 6.1 with support for OCaml 4.10.0 has been released yesterday. This PR is ready to be reviewed and merged.

kit-ty-kate:
I'm not sure who has commit rights on this repository. cc @whitequark if you have some time? (I can take care of the release on the opam side if a tag is available)

gasche:
I have commit rights and "prepare a ppx_deriving release" has been on my TODO list every day this week. I hope to take care of it... soon?

gasche:
(My plan is to do a minor release with this PR and a major release with ppxlib support.)

gasche:
I did a minor refactoring of the 4.10-support patch, to avoid duplication of code logic in the 4.10 and non-4.10 cases. I'm now waiting for the CI results.

gasche:
release-preparation PR on opam-repository for revdep testing: https://github.com/ocaml/opam-repository/pull/15872

kit-ty-kate:
The sematics of your refacting is not the same as my original patch. Are you sure this is the right way? `module_nesting` might be used in `f ()` and it was reset to the old value after the function call

kit-ty-kate:
If you're sure this is the right way, I'll go ahead and launch a revdeps check myself on 4.09 and 4.10 (opam-repository CI does not check for 4.10 so this patch wasn't checked there)

gasche:
I see your point. I was not fond of the duplication in your patch, but now I think that the best approach is to revert to the version that you checked. Let me push back your old commit, plus the `.opam` change.

gasche:
(I think it would be possible to have a commit that always saves `module_nesting` and also avoids the duplication, but it's not worth the hassle.)

gasche:
There, I just reverted to your original version.

kit-ty-kate:
Ok! If you're ok with the change I think it should be ok to merge and and release now.

gasche:
I merged and tagged; new opam-repository PR: https://github.com/ocaml/opam-repository/pull/15875

PR #220, opened 2020-05-01
OCaml 4.11 support

kit-ty-kate:
Relies on https://github.com/ocaml-ppx/ppx_tools/pull/82

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,6 @@ env:
   - OCAML_VERSION=4.08
   - OCAML_VERSION=4.09
   - OCAML_VERSION=4.10
+  - OCAML_VERSION=4.11.0+trunk OCAML_BETA=enable
 os:
   - linux


thierry-martinez:
Thank you, @kit-ty-kate ! There is a long standing pull request https://github.com/ocaml-ppx/ppx_deriving/pull/210 , which is about to be accepted (I think!), and which migrates `ppx_deriving` to `ppxlib`, dropping the dependency to `ppx_tools`. However, the compatibility with OCaml 4.11.0 has to be checked.

kit-ty-kate:
I don't think #210 changes anything to the core of this PR. I'll happily rebase it if it does though.
By `relies on the ppx_tools PR` I meant more that currently it requires that PR to be merged and released first to be able to compile it in CI.

For the moment this PR is a draft as no alpha/beta release of OCaml 4.11 is available yet, so maybe the API can still change.

kit-ty-kate:
The switch to ppxlib seems to mean this PR is no longer needed. Maybe a the CI entry can still be added?

gasche:
I would be happy to merge the CI support, but for now the 4.11 CI seems to be failing. Do we want to understand why before merging? (I am worried about the "noise" aspect of every CI showing red from the big picture before this is fixed.)

thierry-martinez:
CI is green now! Merged. Thank you very much!

PR #222, opened 2020-05-22
Tentative lightning update for OCaml 4.11.0 before introducing ppxlib

thierry-martinez:
This pull-request tries to propose a small set of changes relatively to the last release to make an intermediate release (let say 4.5.0) before migrating to `ppxlib`.
It appears that the migration to `ppxlib` cannot be made without breaking backward compatibility, therefore it makes sense to make it a major release, i.e. 5.0.0.
Meanwhile, it can be useful to propose another release, a minor one this time, just to provide compatibility with OCaml 4.11.0, while breaking backward compatibility as few as possible.
This work was done above the work already done by @kit-ty-kate in https://github.com/ocaml-ppx/ppx_deriving/pull/220 . Many thanks to her!


gasche:
I am not convinced that it is correct to move common logic to `ppx_deriving_runtime`. The intent of this module is to expose definitions that *generated code* uses, not helper definitions for the code generators. (Adding a dependency to compilerlibs there feels wrong, for example.) For now I would rather inline the helpers in each module, or add a small library to provide code in `ppx_deriving_main` and `api/ppx_deriving`.

Also: we should have a `pre-ppxlib` branch that this PR would use as target.

thierry-martinez:
@gasche Thanks for suggesting the branch creation, and you are right, the helpers for destructing strings should be in the API (I don't think we need another module for helpers). I propose to keep `Option` compatibility in runtime though, next to `Result`.

gasche:
Thanks! The `string_of_*_opt` operations are very convincing, they fit nicely within the existing API. I am slightly frustrated that api/ppx_deriving now depends on ppx_deriving_runtime, but this is not problematic, and I agree having an Option compatibility layer there is sensible -- it may be helpful to some plugins.

gasche:
This is good to merge if the CI agrees.

gasche:
This looks good to me, assuming the CI is happy. It would be nice to have a `pre-ppxlib` branch here as well.

thierry-martinez:
CI needs that ppx_tools be ported to 4.11 (https://github.com/ocaml-ppx/ppx_tools/pull/82) to succeed.

src/api/ppx_deriving.cppo.ml
```diff
@@ -142,6 +143,44 @@ let create =
...
+    Asttypes.constant
+  #endif
+
+let string_of_constant_opt (constant : constant) : string option =
+  match constant with
+  #if OCAML_VERSION >= (4, 11, 0)
+  | Pconst_string (s, _, _) ->
+  #else
+  | Pconst_string (s, _) ->
+  #endif
```
gasche, line 159: Nitpick: it would be a bit more idiomatic in this file to define `Pconst_string` as a macro (or `Pconst_string_patt` if we have reasons to believe that we would also need it as a term-former) to have the same interface on all versions, at the top of the file (where we have other definitions in this style), and use the macro here.

```ocaml
#if OCAML_VERSION >= (4, 11, 0)
#define Pconst_string(s, loc) Pconst_string(s, loc, _)
#endif
```

gasche:
@thierry-martinez I will not work on this more this afternoon. *If* you get to a point where you are satisfied, please feel free to merge. (You may want to cleanup the history first.)

thierry-martinez:
@gasche Ok, I will do that. Thank you very much for your help!

thierry-martinez:
Finally merged! Sorry for CI golfing!

PR #223, opened 2020-05-23
Use "include module type of struct include" to get all equalities

thierry-martinez:
Some equalities were missing in `Ppx_deriving_runtime`, for instance the equality between `Pervasives.result` and `Result.result`, which leads to some strange type errors (for instance, the one corrected here for OCaml <4.08: https://github.com/ocaml-ppx/ppx_deriving/pull/222/commits/f89e120c8f8fc1cd1e319bf3e8217d8e7e62a923).

Instead of trying to enumerate all equalities explicitly, this PR proposes to use the construction

```ocaml
include module type of struct
  include M
end
```

to precisely get the signature of `include M`.

Moreover, it seems that we do not lose anything by using module aliases in signatures.

### DIFF
--- a/src/runtime/ppx_deriving_runtime.cppo.ml
+++ b/src/runtime/ppx_deriving_runtime.cppo.ml
@@ -14,7 +14,7 @@ type nonrec int64 = int64
 type nonrec 'a lazy_t = 'a lazy_t
 type nonrec bytes = bytes
 
-#if OCAML_VERSION >= (4, 08, 0)
+#if OCAML_VERSION >= (4, 07, 0)
 (* We require 4.08 while 4.07 already has a Stdlib module.
    In 4.07, the type equalities on Stdlib.Pervasives
    are not strong enough for the 'include Stdlib'
@@ -25,11 +25,11 @@ module Stdlib = Stdlib
 include Stdlib
 
 module Result = struct
-  type ('a, 'b) t = ('a, 'b) Result.t =
+  type ('a, 'b) t = ('a, 'b) result =
     | Ok of 'a
     | Error of 'b
 
-  type ('a, 'b) result = ('a, 'b) Result.t =
+  type ('a, 'b) result = ('a, 'b) t =
     | Ok of 'a
     | Error of 'b
 end
--- a/src/runtime/ppx_deriving_runtime.cppo.mli
+++ b/src/runtime/ppx_deriving_runtime.cppo.mli
@@ -22,81 +22,53 @@ type nonrec bytes = bytes
 (** {2 Predefined modules}
     {3 Operations on predefined types} *)
 
-#if OCAML_VERSION >= (4, 08, 0)
-include (module type of Stdlib with
-  type fpclass = Stdlib.fpclass and
-  type in_channel = Stdlib.in_channel and
-  type out_channel = Stdlib.out_channel and
-  type open_flag = Stdlib.open_flag and
-  type 'a ref = 'a Stdlib.ref and
-  type ('a, 'b, 'c, 'd, 'e, 'f) format6 = ('a, 'b, 'c, 'd, 'e, 'f) Stdlib.format6 and
-  type ('a, 'b, 'c, 'd) format4 = ('a, 'b, 'c, 'd) Stdlib.format4 and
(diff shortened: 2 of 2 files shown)


src/runtime/ppx_deriving_runtime.cppo.ml
```diff
@@ -14,7 +14,7 @@ type nonrec int64 = int64
 type nonrec 'a lazy_t = 'a lazy_t
 type nonrec bytes = bytes
 
-#if OCAML_VERSION >= (4, 08, 0)
+#if OCAML_VERSION >= (4, 07, 0)
```
gasche, line 17: I'm slightly worried by this change here; but I'm not sure how we would know if it was incorrect (see the comment below); is it fine if the type-checker agress?

Note that in pre-ppxlib you added Option to the other branch, but not this one. If you do this then you have to include it here as well, as Option in 4.08.

thierry-martinez: (You are right for `Option`: I will cherry-pick into `master` the addition of `Option` from `pre-ppxlib`, and if we agree on this PR, I will cherry-pick it as well into `pre-ppxlib`.)

I understood the comment below (which should go away, indeed, if we agree with this change) that the reason not to handle OCaml 4.07 as its successors was indeed to make the type-checker agree.

If we make `module Stdlib = Stdlib` in the implementation, we cannot make `module Stdlib = Pervasives` in the interface, but for a strange reason, the interface that is used for 4.08 and above does not type check in 4.07, basically because the following signature does not type check either (`out_channel` can be replaced by any type declared in `Stdlib`):
```ocaml
include (module type of Stdlib with
  type out_channel = Stdlib.out_channel)
```
I should admit I didn't track down why this signature does not type check in 4.07 and type check with 4.08 and above. I checked however that 4.07 still has the type equalities `out_channel = Pervasives.out_channel = Stdlib.out_channel = Stdlib.Pervasives.out_channel`.

The problem disappears with `include module type of struct include`, and I don't see which type equalities we could have lost with that construction, so I think it is fine to unify the code for 4.07 with 4.08 and above. But if you prefer to stay on the safe path and keep the former version for 4.07, I would agree with that.

gasche: I don't understand why we need to use `include (module type of (struct include M end))` instead of `include (module type of M)` directly, do you? I remember that these two constructs are not type-checked in the same way, but not what's the difference (something something strengthening), I had the vague feeling that `struct include M end` should be weaker.

thierry-martinez: `module type of struct include M end` gets the equalities generated by `include M`. For instance:
```ocaml
# module M = struct type t end;;
module M : sig type t end
# module type T = module type of M;;
module type T = sig type t end
# module type T = module type of struct include M end;;
module type T = sig type t = M.t end
```
It is shorter but should be equivalent to write all the equalities explicitly:
```ocaml
# module type T = (module type of M with type t = M.t);;
module type T = sig type t = M.t end
```

thierry-martinez: (I mentioned this trick some times ago here: http://gallium.inria.fr/blog/overriding-submodules/#comment-4440770785 -- see the second example in my comment, which is shorter but should be equivalent to the solution you proposed above for overriding `Bigarray`.)

gasche: Ah, so `module type of struct include M end` is the *stronger* variant, not the weaker one. Confusing...

With this precision in, I think that I am happy with the change.

gasche:
Now that I've been reminded a couple arcane details of the OCaml module system, and got the reassurance that @thierry-martinez remembers them even better, I think that the PR is a good idea (it does simplify the code significantly).

I still have a lingering feeling that there is something slightly dangerous about making assumptions on how well advanced aspects of the OCaml type system are going to work on *all* OCaml versions we support. But we have CI support, and if there was a problem with this particular pattern I believe that we would see it that way, and the code simplification really is nice and will make future maintenance easier.

 @thierry-martinez, I would propose that you merge when you are ready.

thierry-martinez:
Merged! Thank you for the review, @gasche!

PR #226, opened 2020-05-24
Fix #183 (pre-ppxlib only): CI testing for our reverse dependencies

### LINKED ISSUE #183: maintenance feature request: CI testing of some of our reverse dependencies
It would be nice if our CI would test the build of some of our reverse depencies (after pinning ppx_deriving), for example a couple plugins and a couple users of plugins. This could help us detecting breakage earlier in the release cycle.

(I don't have the knowledge of the best way to do that in immediate memory, so please feel free to point at an example of other project that does this, or even send a PR. Which reverse dependencies we use may depend on the opam-repository-fixing that we are in the process of doing in preparation for the 4.3 release, so I think we could wait to decide this after the release is out.)

thierry-martinez:
This should implement @gasche suggestion:
https://github.com/ocaml-ppx/ppx_deriving/issues/183

We should do that for `master` as well (I expect there will be
more failures there!) but I begin here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,16 @@ install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.t
 script: bash -ex .travis-opam.sh
 env:
   matrix:
-  - OCAML_VERSION=4.02 # we require >=4.02.2 but this picks 4.02.3
-  - OCAML_VERSION=4.03
-  - OCAML_VERSION=4.04
-  - OCAML_VERSION=4.05
-  - OCAML_VERSION=4.06
-  - OCAML_VERSION=4.07
-  - OCAML_VERSION=4.08
-  - OCAML_VERSION=4.09
-  - OCAML_VERSION=4.10
-  - OCAML_VERSION=4.11.0+trunk OCAML_BETA=enable
+  - REVDEPS=y OCAML_VERSION=4.02 # we require >=4.02.2 but this picks 4.02.3
+  - REVDEPS=y OCAML_VERSION=4.03
+  - REVDEPS=y OCAML_VERSION=4.04
+  - REVDEPS=y OCAML_VERSION=4.05
+  - REVDEPS=y OCAML_VERSION=4.06
+  - REVDEPS=y OCAML_VERSION=4.07
+  - REVDEPS=y OCAML_VERSION=4.08
+  - REVDEPS=y OCAML_VERSION=4.09
+  - REVDEPS=y OCAML_VERSION=4.10
+  - REVDEPS=y OCAML_VERSION=4.11.0+trunk OCAML_BETA=enable
     PINS="ppx_tools:https://github.com/kit-ty-kate/ppx_tools.git#411"
 os:
   - linux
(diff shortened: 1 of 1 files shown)


gasche:
I'm not completely sure how `REVDEPS=y` works (I've never used it myself) but I suspect that it will not work well for us: `ppx_deriving` has a *lot* of reverse dependencies, and Travis is probably going to kill our job before it completes.

The way I usually test for reverse dependencies is by submitting a draft PR on opam-repository, and letting the excellent opam-repository CI do it for me. (There is a bit of ceremony as it requires creating a fake opam file during the pre-release.) See [this PR for example](https://github.com/ocaml/opam-repository/pull/14259) (in that case it was merged before I could force-update the URL to point to the final release, so you can see exactly what I submitted on the first try. Usually I force-update once I'm satisfied with revdeps testing, so the PR becomes the final PR.)

Would you try to do it, or should I take care of this process?

thierry-martinez:
I will try to do that (in fact, I planned to do that first, and then I saw your #183 and thought that it could be nice to have the check here). Do you know how opam-repository CI does its reverse-dependency check not to have the same problem as with `REVDEPS=y`?

gasche:
I don't know how REVDEPS=y is implemented, but I know that the opam-repository CI is running on their own cloud infrastructure (not a Travis-provided virtual machine), and they use a lot of compute power for it.

gasche:
Ah, I think the intent of #183 is to have just a few hand-listed dependencies (typically: some plugins, hopefully all, and a couple users of those plugins, or at least running the plugin testsuite). This might work in terms of CI time.

gasche:
The source of REVDEPS logic is at https://github.com/ocaml/ocaml-ci-scripts/blob/master/src/ci_opam.ml#L344

It runs `opam list --depends-on <package> --short` and then tries to `opam install` each of those in sequence. For us this means more than 150 packages; I don't think Travis can do this in a reasonable time limit, and there is a risk that it blocks your CI-golfing in other PRs in the meantime.

thierry-martinez:
I cancelled the build. Actually, I saw to the source of REVDEPS logic because it is how I learnt about this environment variable: I saw that opam-repository CI used `--depends-on` and I grepped it in `opam-ci`. So, a good solution appears to be:

- to prepare a draft PR in opam-repository to check all the rev-dependencies,

- to find how to check only some hand-listed rev-dependencies here.

I will try to do that now! Sorry for all these trials-and-errors!

thierry-martinez:
I just made the PR https://github.com/ocaml/opam-repository/pull/16511 : thanks for the example, @gasche !
I also just committed a hand-picked list of dependencies, but I don't understand why Travis doesn't try to build it. Is it because I cancelled the previous build? Anyway, I will try to close the PR and re-open it again.

thierry-martinez:
Let's see if Travis will try to build the PR this time.

thierry-martinez:
Nope. Should I open a new PR? I will do that, except if someone knows a better solution. @gasche, sorry to importune you again, but perhaps you have some lights on this.

gasche:
No, I don't know; Travis is mostly a blackbox.

Thanks for your efforts improving the CI testing!

PR #231, opened 2020-09-18
Port to ppxlib 0.16 / ocaml-migrate-parsetree 2.0.0

thierry-martinez:
This PR is an attempt to port `ppx_deriving` to `ppxlib 0.16`.


gasche:
(Let me explicitate some context that is shared with @thierry-martinez, but may help other people finding the PR, or ourselves in the future.) In the previous implementation, we were careful to use Ppxlib but without using the OCaml AST version it encourages, rather using the "current" version whenever possible -- using ocaml-migrate-parsetree to transform from one to another. This choice was made to avoid a difficulty coming from Ppxlib's non-versioned design, where users are "stuck" with one version of OCaml that may be lagging a few releases behind the current OCaml release.

In recent months, ppxlib maintainers have taken step to reduce the lag, and announced their intention to follow closely new OCaml releases. This means that we will have to update more often (if we want to follow recent ppxlib releases), but our users will not pay the cost of the "lag", even if we use Ppxlib's AST directly instead of the "current" AST.

This PR applies this design change. This simplifies the implementation quite a bit, which is excellent news.

Another change in this PR is a removal of the driver-registration logic, which was done using ocaml-migrate-parsetree parser facilities. The ppxlib maintainers (who also maintain ocaml-migrate-parsetree) have decided to consolidate all driver logic in ppxlib, so we cannot do a non-ppxlib registration anymore, and we can again simplify the implementation by removing this feature.

In terms of code, this PR thus has very nice properties. I'm not completely sure that the change preserves compatibility for plugin authors, or whether we will have to upgrade plugins again.

thierry-martinez:
Thanks, @gasche, for the recap. I tried to add `REVDEPS` to the CI: the result is not good yet, I will try to improve that tomorrow!

kit-ty-kate:
Superseded by #237

Issue #233, opened 2020-10-20
[proposal] ppx_deriving 5.0 and 5.1


kit-ty-kate:
OCaml 4.12 just entered its freezing period. Last release we decided to wait a bit to release ppx_deriving with ppxlib support (aka ppx_deriving 5.0) to leave people some time to adjust to the new interface.

This time I would encourage to release it as soon as possible as-is (without https://github.com/ocaml-ppx/ppx_deriving/pull/231), so that people get some incentive to get up-to-date then release ppx_deriving 5.1 which would include https://github.com/ocaml-ppx/ppx_deriving/pull/231.
This way packages down the line that still require ppxlib < 0.16.0 can still be compatible with some versions of ppx_deriving 5.x, this would increase quick adoption somewhat.

@gasche @thierry-martinez I can handle the release if you don't have the time.

gasche:
Your plan sounds good to me and, yes, for me it would be very useful if you could handle the release. Thanks a lot.

 Are you planning to use opam-health-check (or whatever the new name is :) to assess breakage when we first release the ppxlib support? (I don't remember if you already did this, apologies in this case.) If so, will you publish the results here?

kit-ty-kate:
I was more planning to just use the regular opam-repository CI for this as many sub-libraries already constrain ppx_deriving to < 5.0, but since you asked I launched a special check. Here are the results:

Packages incompatible with the new API:
* visitors
* ppx_deriving_madcast
* js_of_ocaml-ppx_deriving_json < 3.5.0 (from 3.5.0 it doesn't use ppx_deriving anymore)
* ppx_deriving_rpc < 6.0.0 (from 6.0.0 it doesn't use ppx_deriving anymore)

Packages also incompatible (use ppx_tools directly and forgot to set the dependency explicitly):
* lens

Packages that fail because they use `seq` but ppx_deriving now use `stdlib-shims`, and those two have clashing interfaces (OCaml < 4.07 only):
* jwto
* gobba
* icalendar
* minicaml
* mirage-nat
* mmdb
* nbd
* owork
* protocol-9p-unix (< 0.12)
* tls (< 0.12)
* user-agent-parser

At the moment, I do not have the rights on this repo to make the release. Would it be possible to get it?

gasche:
Thanks a lot for the checking. Comments:

- I don't think there is any problem with the packages that switched to ppxlib directly (ppx_deriving_rpc and jsoo-ppx_deriving_json)
[...]

PR #238, opened 2020-10-27
Release 5.2

kit-ty-kate:
@thierry-martinez @gasche was there anything else to do in https://github.com/ocaml-ppx/ppx_deriving/pull/231? Do I need to elaborate on something on the changelog?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+5.2 (27/10/2020)
+----------------
+
+* Update to ppxlib 0.16.0 #237
+  (Thierry Martinez, Gabriel Scherer, Kate Deplaix)
+
 5.1 (26/10/2020)
 ----------------
 


gasche:
On my side, I have nothing more to suggest. Thanks for the work!

In https://github.com/ocaml-ppx/ppx_deriving/issues/233#issuecomment-714605717 you posted some opam CI results; there remains three failures of our dependencies (visitors, ppx_deriving_madcast, and lens) that I think we should ideally do something about -- at least warn the maintainers, at best come up with a fix. I don't know if this is something you would be interested in helping with. If not, just let me know, and I will try to take care of it myself.

kit-ty-kate:
> On my side, I have nothing more to suggest. Thanks for the work!
> 
> In [#233 (comment)](https://github.com/ocaml-ppx/ppx_deriving/issues/233#issuecomment-714605717) you posted some opam CI results; there remains three failures of our dependencies (visitors, ppx_deriving_madcast, and lens) that I think we should ideally do something about -- at least warn the maintainers, at best come up with a fix. I don't know if this is something you would be interested in helping with. If not, just let me know, and I will try to take care of it myself.

Don't worry, I'm in charge of that. I just want to finish the ppx_deriving releases first then I'll do the rest. It'll show up in my next run of opam-health-check for 4.12 anyway so I'll *have* to do it ^^

PR #240, opened 2020-10-27
Upgrade and clean up legacy code

kit-ty-kate:
This PR cleans up several things:
* Removes direct dependencies to `result` and `ocaml-migrate-parsetree`. Both were unused.
* Upgrade from `ounit` to the `ounit2` package
* Upgrade to dune 2.7 (which will generate the opam file for us)
* Enable standard dune warnings (+ `-w -9` otherwise it would have taken me days to do this). This allowed me to detect several unused code (notably in the tests), variables and open.

PR #246, opened 2020-11-25
(almost) use the set of standard flags from dune

kit-ty-kate:
Supersedes #240


gasche (approved):
Obviously an improvement. Thanks!

PR #248, opened 2020-12-11
Fix some of Ast_convenience's functions when giving a full path ident (e.g. M.ident, not just ident)

kit-ty-kate:
This is required for `ppx_deriving_protobuf` to work with minimal changes. See: https://github.com/ocaml-ppx/ppx_deriving_protobuf/pull/39


kit-ty-kate:
Reasoning for this change given in https://github.com/ocaml-ppx/ppx_deriving_protobuf/pull/39#issuecomment-771244270

PR #255, opened 2021-12-01
chore: remove artifact

rgrinberg:
compat_macros.cppo doesn't exist

@kit-ty-kate there's no bug, but it's confusing to keep this


gasche:
Thanks! This is a leftover from the pre-ppxlib days I guess.

PR #256, opened 2022-01-28
Fix ppx_deriving_fold for polymorphic variants

danielmercier:


sim642 (approved):
Thanks!
I rebased and also added a test.

PR #258, opened 2022-02-16
Add OCaml 5.00 support to the tests and update documentation

kit-ty-kate:


src_test/map/test_deriving_map.cppo.ml
```diff
@@ -99,7 +99,7 @@ let test_var2 ctxt =
   assert_equal ~printer:(show_var2 fmt_int) (A2 5) (map_var2 int_of_float (A2 5.))
 
 let test_var3 ctxt =
-  let show,map = show_var3 fmt_int fmt_str, map_var3 ((+)1) String.uppercase [@warning "-3"] in
+  let show,map = show_var3 fmt_int fmt_str, map_var3 ((+)1) String.uppercase_ascii in
```
gasche, line 102: Out of curiosity: I assume you checked that we don't support anymore the old OCaml versions without `uppercase_ascii`?

gasche (commented):
Looks very sensible, thanks!

src_test/map/test_deriving_map.cppo.ml
kit-ty-kate: Yes, `uppercase_ascii` was added in OCaml 4.03 which we don’t support anyway

PR #267, opened 2023-03-18
show: handle [@printer] in polymorphic variants

ghuysmans:


brycenichols:
Code generated by openapi-generator-cli for ocaml uses custom printers for polymorphic variant branches as a way of supporting jsonschema enum types in generated client library code.  I would have expected (and the openapi-generator-cli authors did, too)  that the annotations would work the same way as for regular variants, but instead, the custom printers are silently ignored.  The fix in this PR would be very helpful.

src_plugins/show/ppx_deriving_show.cppo.ml
```diff
@@ -173,20 +173,31 @@ let rec expr_of_typ quoter typ =
...
+          match attr_printer field.prf_attributes, field.prf_desc with
+          | None, Rtag(label, true (*empty*), []) ->
             let label = label.txt in
             Exp.case (Pat.variant label None)
                      [%expr Ppx_deriving_runtime.Format.pp_print_string fmt [%e str ("`" ^ label)]]
-          | Rtag(label, false, [typ]) ->
+          | Some printer, Rtag(label, true (*empty*), []) ->
+            let label = label.txt in
+            Exp.case (Pat.variant label None)
+                     [%expr [%e printer] fmt ()]
```
gasche, line 184: I think it would be easier to factorize the two cases better, the code is identical except for the second argument of `Exp.case` that depends on the optional printer attribute.

gasche, line 184: I am worried about the use of `printer` directly when all other such patterns in the code use `wrap_printer quoter printer`. I don't remember the details of how this works, but I would follow the rest of the code here.

src_plugins/show/ppx_deriving_show.cppo.ml
```diff
@@ -173,20 +173,31 @@ let rec expr_of_typ quoter typ =
...
             let label = label.txt in
             Exp.case (Pat.variant label (Some [%pat? x]))
                      [%expr Ppx_deriving_runtime.Format.fprintf fmt [%e str ("`" ^ label ^ " (@[<hov>")];
                             [%e expr_of_typ typ] x;
                             Ppx_deriving_runtime.Format.fprintf fmt "@])"]
-          | Rinherit({ ptyp_desc = Ptyp_constr (tname, _) } as typ) ->
+          | Some printer, Rtag(label, false, [typ]) ->
+            let label = label.txt in
+            Exp.case (Pat.variant label (Some [%pat? x]))
+                     [%expr [%e printer] fmt x]
```
gasche, line 194: Same remarks: you could share more code (but not doing so is okay), and `printer` should be wrapped.

gasche (commented):
Thanks for the ping! I reviewed the PR. There is a comment on code style, which is optional, but the lack of `wrap_printer` is probably a bug.

@ghuysmans apologies for the delay in reviewing this. Would you by chance be able to tell if you want to take care of your PR despite the long wait, or if you have moved to other things and would rather have someone else adopt it?

sim642:
Closing in favor of #286.

brycenichols:
Happy to see a resolution here!

Issue #272, opened 2024-01-16
`ppx_deriving.make` breaks on mutually recurive types


shonfeder:
In `ex.ml`

```ocaml
module Ex = struct
  type t = {a : int; b: s}
  [@@deriving make]
  and s = t list
end
```

With the `dune` file:

```scheme
(library
 (name ex)
 (modules ex)
 (preprocess (pps ppx_deriving.make)))
```

`dune build` gives

```sh
$ dune build                    
File "ex.ml", line 5, characters 2-16:
5 |   and s = t list
      ^^^^^^^^^^^^^^
Error: make can be derived only for record types
```

If we put another deriving annotation on `s` and swap the order the types are declared, the compilation goes thru but the `make` annotation is ignored:

```ocaml
module Ex = struct
  type s = t list
  [@@deriving show]

  and t = {a : int; b: s}
  [@@deriving make]
end
```

Which produces

```ocaml
[@@@ocaml.ppx.context
  {
    tool_name = "ppx_driver";
    include_dirs = [];
    load_path = [];
    open_modules = [];
    for_package = None;
    debug = false;
    use_threads = false;
    use_vmthreads = false;
    recursive_types = false;
    principal = false;
    transparent_modules = false;
    unboxed_types = false;
    unsafe_string = false;
    cookies = [("library-name", "ex")]
  }]
module Ex =
  struct
    type s = t list[@@deriving show]
    and t = {
      a: int ;
      b: s }[@@deriving make]
    let rec pp_s :
      Ppx_deriving_runtime.Format.formatter -> s -> Ppx_deriving_runtime.unit
      =
      let __0 () = pp in
      ((let open! ((Ppx_deriving_runtime)[@ocaml.warning "-A"]) in
          fun fmt ->
            fun x ->
              Ppx_deriving_runtime.Format.fprintf fmt "@[<2>[";
              ignore
                (List.fold_left
                   (fun sep ->
                      fun x ->
                        if sep
                        then Ppx_deriving_runtime.Format.fprintf fmt ";@ ";
                        ((__0 ()) fmt) x;
                        true) false x);
              Ppx_deriving_runtime.Format.fprintf fmt "@,]@]")
        [@ocaml.warning "-A"])
    and show_s : s -> Ppx_deriving_runtime.string =
      fun x -> Ppx_deriving_runtime.Format.asprintf "%a" pp_s x[@@ocaml.warning
                                                                 "-32"]
    and pp :
[...]

PR #273, opened 2024-02-24
Regression test for incorrect behavior when deriving on mutually recursive types

shonfeder:
Adds a regression test for #272

Note that is this on top of #263 and only 7dbaf97f797d127cd6d23aa11863e480febac289 is relevant currently.

During the ppxlib dev meeting, @pitag-ha and @NathanReb asked me to see if #263 would fix the problem here. The failing test here unfortunately demonstrates that it does not fix the problem as is:

```
File "src_test/make/test_deriving_make.ml", lines 63-64, characters 2-19:
63 | ..and secondary_recursive_type = string
64 |   [@@deriving show]
Error: make can be derived only for record types
```

See https://ocaml.ci.dev/github/ocaml-ppx/ppx_deriving/commit/7dbaf97f797d127cd6d23aa11863e480febac289/variant/alpine-3.19-4.14_opam-2.1#L238-241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 (unreleased)
 ------------
 
+* Port standard plugins to ppxlib registration and attributes
+  #263
+  (Simmo Saan)
+
 * Introduce `Ppx_deriving_runtime.Stdlib` with OCaml >= 4.07. This module
   already exists in OCaml < 4.07 but was missing otherwise.
 
--- a/ppx_deriving.opam
+++ b/ppx_deriving.opam
@@ -19,7 +19,7 @@ depends: [
   "cppo" {build}
   "ocamlfind"
   "ppx_derivers"
-  "ppxlib" {>= "0.27.0"}
+  "ppxlib" {>= "0.29.0"}
   "result"
   "ounit2" {with-test}
 ]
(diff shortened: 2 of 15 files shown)


NathanReb:
Just to clarify what the bug is and what is the expected fix.

What happens here is that it tries to derive make for all types which it can't. What you would like instead is that it would only derive make for the type declarations in the recursive type decl that have the [@@deriving make] right?

NathanReb:
It would be good to have expect tests there instead, that way we could have merged this PR with a green CI and update the test along with the bug fix.

I'll look into merging #263 once the 5.02 compat in ppxlib is dealt with and will come back to this afterward! Sorry for the delay!

shonfeder:
> Just to clarify what the bug is and what is the expected fix.
> 
> 
> 
> What happens here is that it tries to derive make for all types which it can't. What you would like instead is that it would only derive make for the type declarations in the recursive type decl that have the [@@deriving make] right?

That's right. It's deriving for declarations which are not annotated, which then breaks when these are underivable. Instead, it should only derive for declarations that are annotated.

shonfeder:
> It would be good to have expect tests there instead, that way we could have merged this PR with a green CI and update the test along with the bug fix.
> 
> 
> 
> I'll look into merging #263 once the 5.02 compat in ppxlib is dealt with and will come back to this afterward! Sorry for the delay!

No worries! I am not blocked by this.

NathanReb:
I'm closing this as I added your test as part of #281 !

Thanks reporting and reproducing this bug!

PR #277, opened 2024-03-12
Declare `show`'s `with_path` argument for interface deriver as well

NathanReb:
Even though the argument is unused in `.mli` files as it only affects the implementation and not the signature of the function, it used to be accepted back when declared via `ppx_deriving.api` so we need to keep it that way for compatibility.


sim642:
Heh, I also started a fix: https://github.com/ocaml-ppx/ppx_deriving/commit/8d6b749995cb6be145b5598cad9183df2b7f2168. Turns out one needs to bypass the value restriction to get it to compile.

NathanReb:
Argh yeah just realized that! I pushed a fix just now!

NathanReb:
@gasche do you want to give this one a look or should I go ahead and merge?

kit-ty-kate:
If it fixes the packages in opam-repository I think you can just merge it

sim642 (approved):

PR #279, opened 2024-03-12
drop the `result` compatibility package dependency

anmonteiro:
- even though the `Result` module was only added in OCaml 4.08:
  - the type `{Pervasives,Stdlib}.result` has been added as far back as OCaml 4.03 
  - ppx_deriving's lower bound is OCaml 4.05, so it should be fine to remove this dependency

### DIFF
--- a/src_plugins/eq/ppx_deriving_eq.ml
+++ b/src_plugins/eq/ppx_deriving_eq.ml
@@ -94,11 +94,11 @@ and expr_of_typ quoter typ =
           | Some a, Some b -> [%e expr_of_typ typ] a b
           | _ -> false]
       | true, ([%type: ([%t? ok_t], [%t? err_t]) result] |
-               [%type: ([%t? ok_t], [%t? err_t]) Result.result]) ->
+               [%type: ([%t? ok_t], [%t? err_t]) result]) ->
         [%expr fun x y ->
           match x, y with
-          | Result.Ok a, Result.Ok b -> [%e expr_of_typ ok_t] a b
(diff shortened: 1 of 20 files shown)


src_plugins/eq/ppx_deriving_eq.ml
```diff
@@ -94,11 +94,11 @@ and expr_of_typ quoter typ =
           | Some a, Some b -> [%e expr_of_typ typ] a b
           | _ -> false]
       | true, ([%type: ([%t? ok_t], [%t? err_t]) result] |
-               [%type: ([%t? ok_t], [%t? err_t]) Result.result]) ->
+               [%type: ([%t? ok_t], [%t? err_t]) result]) ->
```
sim642, line 97: The two cases are now the same. Same in a bunch of other plugins.

But I wonder what implications this change has because this is a fully syntactic pattern. Code which previously derived explicitly for `Result.result` would no longer match here and fail to derive, no?

sim642 (commented):
I wonder what this means #254 and #260.

src_plugins/eq/ppx_deriving_eq.ml
anmonteiro: that's fair. I didn't catch this as it was mostly find / replace. I think it might probably be warranted to keep this case here, which is just matching on an ident `Result.result`. Since that doesn't re-introduce back the result dependency I think it should be reverted.

NathanReb (commented):
This looks good! I'll fix @sim642 suggestion and merge if that's fine by you @anmonteiro.

PR #281, opened 2024-04-08
Fix a bug with [@@deriving make] on type declarations sets

### LINKED ISSUE #272: `ppx_deriving.make` breaks on mutually recurive types

NathanReb:
Fixes #272.

There was a bug with `[@@deriving make]` that caused errors if it was used on a set of type declarations, recursive or non recursive, that contained a non record type.

The deriver was applied to all type declarations in the set, regardless of where the attributes was attached and failed on the ones that weren't record types.

There is no particularly nice way to handle this in ppxlib, I assume the design decision was that one would only attach a single `[@@deriving ...]` attribute per set and the individual derivers would be responsible to decide which type declarations are relevant.
The driver does allow attaching multiple `[@@deriving ...]` attributes in the same set but performs a merge of all of them so it can lead to confusing situations for the end user.

With all that in mind, there is currently no proper way to tell to which individual type declarations the deriver was attached and even though there might be a hacky one, I'm not sure it would be desirable to use it as it's unlikely that other derivers do so and we probably don't want to have `ppx_deriving.make` behave differently to the rest of the ppx universe in that regard.

The fix will derive `make` for all record types in the set if there is at least one, or will report errors about a misuse of the
deriver otherwise.

The PR contains the initial regression test added in #273 by @shonfeder along with a small refactoring to properly embed errors. This allows us to treat regular errors, for example errors coming from misuse of `[@main]` or `[@split]`, and the `make can only be derived for record types` error differently which we use for the actual fix.

I also updated the README to document how to use `ppx_deriving.make` with type declaration sets.

### DIFF
--- a/src_plugins/make/ppx_deriving_make.ml
+++ b/src_plugins/make/ppx_deriving_make.ml
@@ -11,126 +11,162 @@ let attr_default context = Attribute.declare "deriving.make.default" context
   Ast_pattern.(single_expr_payload __) (fun e -> e)
 let attr_default = (attr_default Attribute.Context.label_declaration, attr_default Attribute.Context.core_type)
 
-let attr_split context = Attribute.declare_flag "deriving.make.split" context
-let ct_attr_split = attr_split Attribute.Context.core_type
-let label_attr_split = attr_split Attribute.Context.label_declaration
+let mk_attr_split context = Attribute.declare_flag "deriving.make.split" context
+let ct_attr_split = mk_attr_split Attribute.Context.core_type
+let label_attr_split = mk_attr_split Attribute.Context.label_declaration
 
-let attr_main context = Attribute.declare_flag "deriving.make.main" context
-let ct_attr_main = attr_main Attribute.Context.core_type
-let label_attr_main = attr_main Attribute.Context.label_declaration
+let attr_split = (label_attr_split, ct_attr_split)
+
+let mk_attr_main context = Attribute.declare_flag "deriving.make.main" context
+let ct_attr_main = mk_attr_main Attribute.Context.core_type
+let label_attr_main = mk_attr_main Attribute.Context.label_declaration
+
+let attr_main = (label_attr_main, ct_attr_main)
 
 let get_label_attribute (label_attr, ct_attr) label =
   match Attribute.get label_attr label with
   | Some _ as v -> v
   | None -> Attribute.get ct_attr label.pld_type
 
+let has_label_flag (label_flag, ct_flag) ({pld_type; _} as label) =
+  Attribute.has_flag ct_flag pld_type || Attribute.has_flag label_flag label
+
 let find_main labels =
-  List.fold_left (fun (main, labels) ({ pld_type; pld_loc; pld_attributes } as label) ->
-    if Attribute.has_flag ct_attr_main pld_type || Attribute.has_flag label_attr_main label then
-      match main with
-      | Some _ -> raise_errorf ~loc:pld_loc "Duplicate [@deriving.%s.main] annotation" deriver
-      | None -> Some label, labels
-    else
-      main, label :: labels)
-    (None, []) labels
+  let mains, regulars = List.partition (has_label_flag attr_main) labels in
+  match mains, regulars with
+  | [], labels -> Ok (None, labels)
+  | [main], labels -> Ok (Some main, labels)
+  | _::{pld_loc; _}::_ , _ ->
+    Error
+      (Location.error_extensionf ~loc:pld_loc
+         "Duplicate [@deriving.%s.main] annotation" deriver)
 
-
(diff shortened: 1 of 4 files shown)


NathanReb:
@sim642 I'm pinging you in case you're interested in reviewing this. I'd recommend the per-commit view here, although I'm happy to split out the error handling bits if it makes the review easier.

I was hoping to get the fix in before hitting a new `6.0` release.

src_plugins/make/ppx_deriving_make.ml
```diff
@@ -11,126 +11,162 @@ let attr_default context = Attribute.declare "deriving.make.default" context
...
+        when name.[String.length name - 1] = 's' ->
+        let name' = String.sub name 0 (String.length name - 1) in
+        Typ.arrow (Label.labelled name') lhs
+          (Typ.arrow (Label.optional name)
+             (wrap_predef_option [%type: [%t rhs] list]) accum)
+      | _ ->
+        Ast_builder.Default.ptyp_extension ~loc
+          (Location.error_extensionf ~loc
+             "[@deriving.%s.split] annotation requires a type of form \
+              'a * 'b list and label name ending with `s'"
```
sim642, line 137: Should the `'b` also be changed here?

NathanReb: That's right, good catch!

PR #286, opened 2024-06-01
 Add show `@printer` support for polymorphic variants 

sim642:
This missing support is listed as one of the motivations for [ppx_deriving_variant_string](https://github.com/ahrefs/ppx_deriving_variant_string?tab=readme-ov-file#why-not-ppx_deriving-show):

> `printer` is only supported in polyvars for some reason, but not on regular variants, which was a feature we wanted to have

I quickly implemented this and only then discovered that #267 already exists.
The difference is that this is for the migrated ppxlib attributes and does use `wrap_printer`.


src_plugins/show/ppx_deriving_show.ml
```diff
@@ -23,6 +23,7 @@ let attr_printer context = Attribute.declare "deriving.show.printer" context
   Ast_pattern.(single_expr_payload __) (fun e -> e)
 let ct_attr_printer = attr_printer Attribute.Context.core_type
 let constr_attr_printer = attr_printer Attribute.Context.constructor_declaration
+let rtag_attr_printer = attr_printer Attribute.Context.rtag
```
gasche, line 26: I was lost for a while as to what `rtag` means. The answer is that this obscure name comes from the parsetree, and it means "a constructor declaration in a polymorphic variant type" (`r` means `row` and `tag` is supposed to suggest `constructor`, I guess). The other PR #267 uses `attr_printer field.prf_attributes` directly, and I suppose that usig the ppxlib mechanism as in here is better.

gasche (approved):
Thanks! This is a nice thing to have, and the implementation in this PR is slightly better than #267. Let's not make it wait as long...

Could you maybe include a CHANGELOG entry to credit both you and @ghuysmans, so that we don't forget to do so later?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+unreleased
+----------
+
+* Add show @printer support for polymorphic variants
+  #286
+  (Simmo Saan and @ghuysmans)
```
gasche, line 6: Guillaume Huysmans (according to git commit data)

PR #288, opened 2024-09-26
Add OCaml 5.3 support

kit-ty-kate:
cc @Octachron


NathanReb (approved):
Thanks for taking care of that!

PR #290, opened 2024-10-16
Deriving show for GADTs (wip)

keigoi:
(TBA)

### DIFF
--- a/src/api/ppx_deriving.cppo.ml
+++ b/src/api/ppx_deriving.cppo.ml
@@ -519,13 +519,22 @@ let poly_arrow_of_type_decl fn type_decl typ =
     let name = name.txt in
     Typ.arrow Label.nolabel (fn (Typ.var name)) typ) type_decl typ
 
+let newtype_arrow_of_type_decl fn type_decl typ =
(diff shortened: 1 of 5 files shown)


gasche:
This looks like impressive work, but it's also not so clear what it does right now from the code. Some early comments:

- When you decide to un-draft this PR, it would help if you could show a simple example of type and generated printing function in the PR description. This helps build intuition when reading the code.

- From a distance I wonder if the logic could be simplified a bit if you allowed yourself to cut some corners, for example:
   + maybe you could generate GADT-like annotations unconditionally (whether the type is actually a GADT or not): `let rec foo : 'a 'b 'c . ....`.
   + you have a clever check to decide which paramaters will be refined as GADTs, but maybe we could do something simpler that is directed by the user: if they use `type ('a, _) t = ...`, use GADT mode for the second parameter. This is different from what we do today, and I wonder if there are scary compatibility issues to think about, for now it's just a thought.

keigoi:
Hello @gasche, thank you for your prompt comments! I filled up the summary now, and any further comments/edits will be highly appreciated. Thanks!

PR #294, opened 2025-03-07
Handle 5.2 AST changes

patricoferris:


patricoferris:
In 5.3 an invariant was added that `Ptyp_poly` does not use the empty list. We must be hitting this now somehow with the release of 5.3 (although I could only recreate it with the 5.2 AST bump?) https://github.com/ocaml/ocaml/pull/11148

Hence in 6ef0b0b we check if the freevars are empty and if so, no `Poly_typ` is added.

NathanReb:
Indeed it was: https://github.com/ocaml-ppx/ppx_deriving_yojson/pull/162

NathanReb (approved):
Looks good to me! I'll cut a release!

PR #297, opened 2025-12-05
ppx_deriving.create: use [Ast_helper.with_default_loc] to derive `create` function with location info

arvidj:
I'm not quite sure what I'm doing, but this seems to fix the following issue for me: https://github.com/ocaml/merlin/issues/1910

Here is a test file:
```ocaml
type b = | A [@@deriving enum]

type t = { foo : string } [@@deriving create, make, eq, ord, show, iter]

let () =
  let (t : t) = create ~foo:"test" () in
  let (t' : t) = make ~foo:"test" in
  let (_ : bool) = equal t t' in
  let (_ : int) = compare t t' in
  let (_ : string) = show t in
  let () = iter t in
  ()
```

With this PR, I can use merlin to jump from the derived functions `create`, `make`, etc, to the type declaration with the deriving attribute.

Without, I would just get errors like: `merlin-call-locate: 'show' seems to originate from '_none_' whose ML file could not be found`.

I've made same fix in all included derivers, perhaps in quite a brutish way. Feedback welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+Unreleased
+-----
+
+* Add location info to `create`, `eq`, `iter`, `make`, `ord`, `show`
+  #297
+  (@arvidj)
+
 6.1.1
 -----
 
--- a/src_plugins/create/ppx_deriving_create.ml
+++ b/src_plugins/create/ppx_deriving_create.ml
@@ -120,9 +120,17 @@ let sig_of_type ({ ptype_loc = loc } as type_decl) =
   [Sig.value (Val.mk (mknoloc (Ppx_deriving.mangle_type_decl (`Prefix deriver) type_decl)) typ)]
 
 let impl_generator = Deriving.Generator.V2.make_noarg (fun ~ctxt:_ (_, type_decls) ->
+  let str_of_type type_decl =
+    Ast_helper.with_default_loc type_decl.ptype_loc @@
+      fun () -> str_of_type type_decl
+  in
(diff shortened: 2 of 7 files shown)


NathanReb (approved):
This looks good to me! Ast_helper's will be deprecated in favor of Ast_builder but this would require rewriting a lot more and fixing the generated code location obviously has higher priority.

Any chance you could add a test?

arvidj:
@NathanReb Thanks for the review!

> Any chance you could add a test?

Sure, but what would you like me to test? That the derived functions still work as expected, or that the location info is now available? Unless I'm mistaken, the existing test suite seems to be mostly about the former. If it's the latter that you want, we could do some tests like those of [bisect_ppx](https://github.com/aantron/bisect_ppx/blob/master/test/instrument/control/for.t) but that might be a bit brittle.

arvidj:
> Sure, but what would you like me to test ...

One option would be to bring in [ocaml-index](https://github.com/ocaml-ppx/ppx_deriving/pull/297) and check that the output contains locations. Or I can try to figure out how `ocaml-index` works and try to replicate that in a test.

arvidj:
> If you're interested in writing it I can provide guidance but I'd understand if that's too much hassle and can add that test separately myself later on.

To be honest I think that is a bit too much for me to take on right now. If you feel it is necessary to merge this PR, then I could work on it, but if not I'd prefer passing :)

NathanReb (approved):
Looks good to me, thanks for the fix!

I'll add tests in a later PR!

PR #301, opened 2026-01-23
Add `[@@deriving pp]` that only generates `pp` and not `show`

### LINKED ISSUE #293: Have you thought about adding `[%derive.pp: <some-type>]` ?
there's `[%show: int]` etc, and that's useful.  But I thought it mgiht be useful to have the same for Format-based pp-functions.  I'm contemplating adding it to `pa_ppx.deriving_plugins.show`, and figured I'd see if you guys wanted to add the same here.

giltho:
Fixes #300 

Disclaimer: AI was used in the writing of this PR, though I have decided on the strategy to implement the changes (following [help in the issue](https://github.com/ocaml-ppx/ppx_deriving/issues/300#issuecomment-3791103697), I have read all code and ensured it made sense, and tested the ppx manually to check that using `[@@deriving pp]` indeed does not generate any `show` function or signature.

Not to _pousser mémé dans les orties_ but I have also spent time re-formatting by hand some of the code I touched, and this codebase could [also](https://ocaml.zulipchat.com/#narrow/channel/527805-compiler-dev/topic/Using.20ocamlformat/with/569719692) use some auto-formatting.


giltho:
(Rebased after the merge of #297)

src_plugins/show/ppx_deriving_show.ml
```diff
@@ -58,15 +63,25 @@ let show_type_of_decl type_decl =
...
+  | Pp_and_show ->
+    let show_sig =
+      Sig.value
+        (Val.mk (mknoloc (Ppx_deriving.mangle_type_decl (`Prefix "show") type_decl))
+                (show_type_of_decl type_decl))
+    in
+    [pp_sig; show_sig]
 
-let rec expr_of_typ quoter typ =
+let rec expr_of_typ ~kind quoter typ =
```
gasche, line 82: Nitpick: It is slightly counterintuitive to take a `kind` parameter here, given that the behavior of the generated code does not depend on this parameter (the code is part of the common printing logic). `kind` is only used to derive the `~deriver:(deriver_name kind)` arguments for error messages -- I agree that showing precise error messages is nice, otherwise users might see errors with a different deriver name than the one they wrote themselves. This could be made slightly clearer by taking a `~deriver` parameter instead of `~kind`.

src_plugins/show/ppx_deriving_show.ml
```diff
@@ -176,26 +191,28 @@ let rec expr_of_typ quoter typ =
...
+                   (Ppx_deriving.string_of_core_type typ)
 
-and expr_of_label_decl quoter { pld_type; pld_attributes } =
+and expr_of_label_decl ~kind quoter { pld_type; pld_attributes } =
   let attrs = pld_type.ptyp_attributes @ pld_attributes in
-  expr_of_typ quoter { pld_type with ptyp_attributes = attrs }
+  expr_of_typ ~kind quoter { pld_type with ptyp_attributes = attrs }
 
-let str_of_type ~with_path ~path ({ ptype_loc = loc } as type_decl) =
+let str_of_type ~kind ~with_path ~path ({ ptype_loc = loc } as type_decl) =
```
gasche, line 209: Nitpick: the code changes to this function are arguably not super-nice (they read like "minimal changes" rather than "what gives the nicer code in the end", because it has code for both derivers and then ends up using only a part of it depending on the `~kind` parameter.

I believe that the current approach is okay, and less work for the author of the PR. But my favorite version would be to have `pp_str_of_type` defined first, and then `show_str_of_type` defined second, and then have `impl_generator` call both or just one as needed. This may fail (to improve readability) if it turns out that there is shared code between the two `*_str_of_type` functions that would be duplicated, but I don't see it right now.

src_plugins/show/ppx_deriving_show.ml
```diff
@@ -299,14 +316,25 @@ let str_of_type ~with_path ~path ({ ptype_loc = loc } as type_decl) =
...
+  in
+  match kind with
+  | Pp_only -> [pp_binding]
+  | Pp_and_show ->
+    let show_binding =
+      Vb.mk ~attrs:[no_warn_32]
+        (Pat.constraint_ show_var show_type)
+        (polymorphize stringprinter)
+    in
+    [pp_binding; show_binding]
```
gasche, line 332: (Micro-nitpick: I think it would be slightly more readable to have `pp_binding` and `show_binding` defined one after another at the same indentation level, and then a trivial `kind`. If you were worried about useless AST computations (I'm not), you could always add a `()` parameter to them to delay computation.

The same pattern occurs in `sig_of_type` above.)

src_plugins/show/ppx_deriving_show.ml
```diff
@@ -328,35 +356,55 @@ let impl_generator = Deriving.Generator.V2.make impl_args (fun ~ctxt (_, type_de
...
   in
-  let str_of_type type_decl =
+    let str_of_type type_decl =
     Ast_helper.with_default_loc type_decl.ptype_loc @@
-      fun () -> str_of_type ~with_path ~path type_decl
+      fun () -> str_of_type ~kind ~with_path ~path type_decl
+  in
+  let rec_flag = match kind with
+    | Pp_only -> Nonrecursive
+    | Pp_and_show -> Recursive
```
gasche, line 365: I don't know why the `rec_flag` computation is like this, and I suspect that it may be wrong. What happens when deriving `pp` for two mutually-recursive types?

src_plugins/show/ppx_deriving_show.ml
```diff
@@ -328,35 +356,55 @@ let impl_generator = Deriving.Generator.V2.make impl_args (fun ~ctxt (_, type_de
...
 
-let deriving: Deriving.t =
+let deriving_pp: Deriving.t =
   Deriving.add
-    deriver
-    ~str_type_decl:impl_generator
-    ~sig_type_decl:intf_generator
+    "pp"
+    ~str_type_decl:(impl_generator Pp_only)
+    ~sig_type_decl:(intf_generator Pp_only)
```
gasche, line 389: Nano-nit: I would actually put the `pp`-only version first, as it is simpler, but I don't mind. (Same with the extensions below.)

src_test/show/test_deriving_show.cppo.ml
```diff
@@ -244,6 +244,39 @@ let test_paths_printer ctxt =
...
+  assert_equal ~printer "(Test_deriving_show.PpBar (1, \"foo\"))"
+    (fmt_to_string pp_pp_only_variant (PpBar (1, "foo")));
+  assert_equal ~printer "{ Test_deriving_show.pp_f1 = 1; pp_f2 = \"bar\" }"
+    (fmt_to_string pp_pp_only_record { pp_f1 = 1; pp_f2 = "bar" });
+  assert_equal ~printer "(42, \"baz\")"
+    (fmt_to_string pp_pp_only_tuple (42, "baz"));
+  assert_equal ~printer "{ Test_deriving_show.pp_v = 99 }"
+    (fmt_to_string (pp_pp_only_param (fun fmt -> Format.fprintf fmt "%d")) { pp_v = 99 })
+
+(* Test [%derive.pp] extension *)
```
gasche, line 269: We don't have a single test for `[%derive.show ]`, could you per chance be motivated to add one?

(Unless I am missing something, offering `[%derive.pp ...]` is a genuinely useful feature that was missing before, so this aspect of your PR adds something more, it does not only offer a way to remove stuff.)

gasche (approved):
This looks good overall, but there is area for improvement if you want to, in varying degrees of (un)importance. I propose that you do what you are inspired to do, possibly nothing (well okay maybe fix the bug), and then I think we could merge.

src_plugins/show/ppx_deriving_show.ml
giltho: You are correct 🙃
Funnily enough this is one of the things I did by hand, not AI, because it was throwing warnings.

giltho:
Can you mark this as `requires changes`? For my own sake. The PR is wrong at least for the rec flag, and I agree with most of the other comments as well (and don't mind applying them all).
Given Caml in the Capital today, I won't have time to change it and I prefer to have this in red when I look at my PR list.

gasche (changes requested):
(see previous review for suggested/required changes)

src_test/show/test_deriving_show.cppo.ml
giltho: Hmm, `[%derive.show]` seems broken, as it calls for types that don't seem to be exposed. I'll have to investigate

giltho: Heh, `derive.show` wasn't broken, it was just that I had written a test with
```ocaml
let show_int_pair : (int * int) -> string = [%derive.show: int * int]
```
but `string` is overriden earlier. I had some error message with some `a8` type that I thought was something internal and shouldn't be exposed, but it's also in the test suite

giltho:
I believe I have addressed all comments.
One dumb question: are deriver transformations ever run in parallel (as in, with shared memory?)

The `~deriver` parameter makes a bit of the code harder to read in my opinion, and I'd only just have a global variable containing the deriver name that is set by the top-level functions once when they start and that's it. Another option would be to use effects but we can't while remaining retro-compatible with OCaml 4

gasche (approved):
I believe that this is good to go if the CI agrees.

(Regarding the `~deriver` parameter: the other options you consider look like they can cause weird surprises and sharp edges, I would rather just keep passing the parameter around.)

giltho:
Yeah, agreed, in the current state this remains the best solution

giltho:
I think the CI errors are unrelated?

sim642:
I fixed the git merge conflicts with #313, #312, #316 and #307, and made a small fix to the deriver name for `[%pp]` errors.

sim642 (approved):

PR #305, opened 2026-06-28
Fix ord incorrectly quoting polymorphic comparison for builtins

sim642:
While inspecting generated code for https://github.com/ocaml/ocaml/issues/14876#issuecomment-4816177881, I noticed something odd: ord is quoting builtins, but eq is not.

The history goes back and forth here.
The quoting was added by #115, which really was trying to fix monomorphization.
Previously, 71cabc00edb48616819ff43517bffc1a3d902552, trying to simplify, removed the monomorphization, which used to be there but without the quoting.
The only reason #115 added the quoting was to get the ord tests to build, but the problem wasn't the lack of quoting but the lack of `[@nobuiltin]` which the corresponding eq test uses.

Although #115 got the ord test to build by adding the quoting, the generated code was subtly wrong. When the "monomorphized" builtin comparison is lifted up by the quoter, the polymorphic comparison is actually applied to the redefined `bool` type, not the standard `bool`. So no monomorphization could actually take place.
Moreover, by performing polymorphic comparison on the redefined type, it completely bypasses the derived `compare_bool`. Since the redefined `bool` type contains a function, polymorphic comparison doesn't even work (i.e., crashes at runtime).

### DIFF
--- a/src_plugins/ord/ppx_deriving_ord.ml
+++ b/src_plugins/ord/ppx_deriving_ord.ml
@@ -77,8 +77,7 @@ and expr_of_typ quoter typ =
               | [%type: int64] | [%type: Int64.t] | [%type: nativeint]
               | [%type: Nativeint.t] | [%type: float] | [%type: bool]
               | [%type: char] | [%type: string] | [%type: bytes]) ->
-        let compare_fn = [%expr fun (a:[%t typ]) b -> Ppx_deriving_runtime.compare a b] in
-        Ppx_deriving.quote ~quoter compare_fn
+        [%expr fun (a:[%t typ]) b -> Ppx_deriving_runtime.compare a b]
       | true, [%type: [%t? typ] ref] ->
         [%expr fun a b -> [%e expr_of_typ typ] !a !b]
       | true, [%type: [%t? typ] list]  ->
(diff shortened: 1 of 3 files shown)


gasche (approved):
Oh, this is a tricky issue indeed! Thanks for catching it.

(I feel out of the loop regarding quoting, I don't remember at all how it works and what its purpose is besides "avoiding hygiene issues", so I'm glad that at least one person, that is, you, has these details clearer in their head.)

gasche:
(cc @kit-ty-kate, @NathanReb) 

I took the liberty of inviting @sim642 to get commit rights, as he evidently would be a competent caretaker for `ppx_deriving`.

(Other topic: I think it would be nice to have a release after this bugfix, but I am not volunteering, I dislike doing releases.)

kit-ty-kate:
Works for me. @sim642 would you like to do the release? I can do it if needed but i'm quite busy with opam and my own few months of backlog now that i'm back from hospital

sim642:
Thanks! I'll give the release a shot.

gasche:
> Thanks!

Well here you are the one doing the work, so it is our place to be grateful. Thanks!

PR #306, opened 2026-07-03
Coalesce syntactic arity in `poly_{fun,apply}_of_type_*`

sim642:
This avoids intermediate closure allocations since OCaml 5.2.

This addresses point 3 from https://github.com/sim642/ppx_deriving_hash/pull/7 more broadly than just ppx_deriving_hash.

### DIFF
--- a/src/api/ppx_deriving.cppo.ml
+++ b/src/api/ppx_deriving.cppo.ml
@@ -514,14 +514,16 @@ let poly_fun_of_type_ext type_ext expr =
     Ast_builder.Default.pexp_fun ~loc Label.nolabel None (pvar ("poly_"^name)) expr) type_ext expr
 
 let poly_apply_of_type_decl type_decl expr =
+  let loc = !Ast_helper.default_loc in
   fold_left_type_decl (fun expr name ->
     let name = name.txt in
-    Exp.apply expr [Label.nolabel, evar ("poly_"^name)]) expr type_decl
+    Ast_builder.Default.pexp_apply ~loc expr [Label.nolabel, evar ("poly_"^name)]) expr type_decl
 
 let poly_apply_of_type_ext type_ext expr =
+  let loc = !Ast_helper.default_loc in
   fold_left_type_ext (fun expr name ->
     let name = name.txt in
-    Exp.apply expr [Label.nolabel, evar ("poly_"^name)]) expr type_ext
+    Ast_builder.Default.pexp_apply ~loc expr [Label.nolabel, evar ("poly_"^name)]) expr type_ext
 
 let poly_arrow_of_type_decl fn type_decl typ =
   fold_right_type_decl (fun name typ ->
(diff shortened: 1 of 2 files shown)


src/api/ppx_deriving.cppo.ml
gasche: Do you know if this is going to generate a single n-ary application, or if it would be working to generate it in one go? (That should be relatively easy.)

src/api/ppx_deriving.cppo.ml
```diff
@@ -504,12 +504,12 @@ let fresh_var bound =
 let poly_fun_of_type_decl type_decl expr =
   fold_right_type_decl (fun name expr ->
     let name = name.txt in
-    Exp.fun_ Label.nolabel None (pvar ("poly_"^name)) expr) type_decl expr
+    Ast_builder.Default.pexp_fun ~loc:Location.none Label.nolabel None (pvar ("poly_"^name)) expr) type_decl expr
```
gasche: Why not use `expr.exp_loc` for the location? (Or `!Ast_helper.default_loc` as the rest of the code?)

(I don't remember the details of how we are supposed to choose locations, maybe these should be the same locations as `expr.pexp_loc` but marked "ghost"?)

In any case, the rest of the file never uses `Location.none`, so these two use-sites you added stand out.

src/api/ppx_deriving.cppo.ml
sim642: This is simple like `Exp.fun_`: it creates n unary applications. Using `Ast_builder.Default.pexp_apply` should do the combining.

I currently didn't touch this because:
1. According to https://github.com/sim642/ppx_deriving_hash/pull/7#issuecomment-4756862382, the nested unary applications don't seem to have any penalty.
2. The eq, ord and hash derivers don't actually use this function. 

But I guess for symmetry it wouldn't hurt to change this either.

src/api/ppx_deriving.cppo.ml
sim642: `Exp.fun_` has `loc` as an optional argument, so I thought it already used none before, but it passes the optional argument deeper where it indeed defaults to `default_loc`.
So I should fix this indeed.

src/api/ppx_deriving.cppo.ml
gasche: In general there definitely are cases where OCaml will handle `(f x y) z` and `(f x y z)` differently, and the compiler will not rewrite from one to the other because it may degrade performance in some corner cases I think. (But: then people ask that `f @@ x @@ y` be rewritten into `f x y`.) If you are careful to coalesce abstractions, I think it would be reasonable to also coalesce applications -- assuming it's not too much work.

gasche (approved):
All my comments have been addressed so I am happy with the current state. Approved.

(I don't know if you intend to simplify/cleanup the history, I will let you decide what to do, and then you can merge yourself.)

PR #307, opened 2026-07-05
Use `Ppxlib.really_recursive` for plugin implementations

sim642:
Currently the implementations are always recursive.
This can be inefficient because recursive functions are compiled differently: https://github.com/ocaml/ocaml/issues/14876#issuecomment-4825973808.

Various Jane Street ppx-s have been doing this for a long time, I guess that's why the function is in ppxlib to begin with.

The subtlety of this approach is that it uses the real recursiveness of the _type declaration_ to choose the recursiveness of the derived value definition. These should match in most cases, although technically this may break some weird uses, e.g. where the type is non-recursive but for whatever reason a deriver override attribute wants to do a recursive call.

### TODO
- [x] Benchmark with Goblint (together with https://github.com/sim642/ppx_deriving_hash/pull/8): No difference on OCaml 4.14, maybe 1% CPU time improvement on OCaml 5.6.0+trunk.

### DIFF
--- a/src_plugins/show/ppx_deriving_show.ml
+++ b/src_plugins/show/ppx_deriving_show.ml
@@ -299,14 +299,14 @@ let str_of_type ~with_path ~path ({ ptype_loc = loc } as type_decl) =
   let show_var =
     pvar (Ppx_deriving.mangle_type_decl (`Prefix "show") type_decl) in
   let no_warn_32 = Ppx_deriving.attr_warning [%expr "-32"] in
-  [Vb.mk (Pat.constraint_ pp_var pp_type)
-         (Ppx_deriving.sanitize ~quoter (polymorphize prettyprinter));
-   Vb.mk ~attrs:[no_warn_32] (Pat.constraint_ show_var show_type) (polymorphize stringprinter);]
+  (Vb.mk (Pat.constraint_ pp_var pp_type)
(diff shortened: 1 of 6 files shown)


src_plugins/show/ppx_deriving_show.ml
```diff
@@ -332,7 +332,10 @@ let impl_generator = Deriving.Generator.V2.make impl_args (fun ~ctxt (_, type_de
     Ast_helper.with_default_loc type_decl.ptype_loc @@
       fun () -> str_of_type ~with_path ~path type_decl
   in
-  [Str.value Recursive (List.concat (List.map str_of_type type_decls))])
+  let (pps, shows) = List.split (List.map str_of_type type_decls) in
+  let rec_flag = really_recursive rec_flag type_decls in
+  [Str.value rec_flag pps;
+   Str.value Nonrecursive shows])
```
sim642, line 339: Splitting the `pp` and `show` functions into separate value definitions instead of one big recursive one (where `show` using `pp` also counts for recursive use, thus no `-39` warning) has one small difference I can think of: `[@printer ...]` attributes cannot refer to the `show` functions being derived in the same group anymore.
But they can still refer to the `pp` functions which would be the more sensible thing to do anyway.

PR #308, opened 2026-07-06
Remove unnecessary cppo from tests

sim642:
This is a continuation of #276.

cppo interferes with ocaml-lsp, making it more annoying to work with the tests, e.g. see the error locations or interactively look at the expanded code.


gasche (approved):
Looks good to me -- removing lines is obviously a win.

Curiosity question: I am not a cppo super-user, but my memory (from using it a couple weeks ago) was that Merlin+Emacs supports it just fine. Is the lack of good support in the IDE a feature specific to ocaml-lsp?

sim642:
> Curiosity question: I am not a cppo super-user, but my memory (from using it a couple weeks ago) was that Merlin+Emacs supports it just fine. Is the lack of good support in the IDE a feature specific to ocaml-lsp?

I didn't investigate what exactly the problem was but it might be due to how cppo is setup via dune in ppx_deriving: dune doesn't support `(preprocess)` with both ppx and cppo at the same time, so the cppo step is a custom rule.

PR #309, opened 2026-07-06
Bump OCaml lower bound to 4.08.0 based on ppxlib 0.36.0 lower bound

sim642:
This is a further continuation of #308.

ppx_deriving has a lower bound on ppxlib 0.36.0, which has a lower bound on OCaml 4.08.0.
So the ppx_deriving lower bound on OCaml 4.05.0 is unachievable and misleading.
Bumping it to 4.08.0 allows a bunch of old cppo code to be removed.


src/api/ppx_deriving.cppo.ml
```diff
@@ -171,7 +171,6 @@ let lookup name =
 let raise_errorf ?sub ?loc fmt =
   let module Location = Ocaml_common.Location in
   let raise_msg str =
-#if OCAML_VERSION >= (4, 08, 0)
```
gasche, line 174: I wonder if there is a simpler way to restructure `raise_errorf` (it's a surprising amount of work just to send an error message), but oh well, the change is obviously correct.

gasche (approved):
Nice simplification, thanks!

src/api/ppx_deriving.cppo.ml
sim642: All the complication here seems to deal with sub-errors. I was hoping that maybe this entire function could be delegated to Ppxlib, but for some reason its version of `raise_errorf` does not have the `sub` argument (perhaps to avoid the same complication?).
And the enum plugin actually uses a sub-error, although the output is strange. I'll open some issues about this.

PR #311, opened 2026-07-06
Remove now-unnecessary unused-rec-flag suppression from `sanitize`

sim642:
This was added in PR #263.
Ppxlib was changed to also return `Nonrecursive` bindings in https://github.com/ocaml-ppx/ppxlib/pull/401, making the warning suppression unnecessary. It was changed in ppxlib 0.30.0, but the current lower bound here is 0.36.0.

Individual plugins still add the suppression to the outer binding themselves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Unreleased
   #306
   (@sim642)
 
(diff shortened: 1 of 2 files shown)


gasche (approved):
Thanks!

PR #312, opened 2026-07-07
Wrap deriver extensions in `with_default_loc`

sim642:
This was done for usual derivers in PR #297, but not for their extension node versions. These also use the quoter, which otherwise has no location.


sim642:
I'll just merge this small change as well so I can do a new release with the many recent fixes soon.

PR #313, opened 2026-07-08
src_plugins: mark default location as loc_ghost

fantazio:
Following https://github.com/ocaml-ppx/ppx_deriving/pull/297, some generated code now have proper locations attached by using the type_decl's location by default instead of a dummy one. However, because the location is used as-is, the generated code is indistinguishable from user code.
This leads to false positives in the dead_code_analyzer (https://github.com/LexiFi/dead_code_analyzer/pull/87).

According to [`Location`'s documentation](https://ocaml.org/manual/5.5/api/compilerlibref/Location.html#TYPEt),
> loc_ghost=false whenever the AST described by the location can be parsed from the location. In all other cases, loc_ghost must be true.
> Most locations produced by the parser have loc_ghost=false.
> When loc_ghost=true, the location is usually a best effort approximation.

The dead_code_analyzer actually relies on this property to decide if an element (value, method, ...) should be tracked or not.

Marking the generated code's default location with `loc_ghost=true` fixes the false positives without breaking the expected merlin behavior mentioned in https://github.com/ocaml-ppx/ppx_deriving/pull/297.

### DIFF
--- a/src_plugins/create/ppx_deriving_create.ml
+++ b/src_plugins/create/ppx_deriving_create.ml
@@ -121,14 +121,14 @@ let sig_of_type ({ ptype_loc = loc } as type_decl) =
 
 let impl_generator = Deriving.Generator.V2.make_noarg (fun ~ctxt:_ (_, type_decls) ->
   let str_of_type type_decl =
-    Ast_helper.with_default_loc type_decl.ptype_loc @@
+    Ast_helper.with_default_loc {type_decl.ptype_loc with loc_ghost = true} @@
       fun () -> str_of_type type_decl
   in
   [Str.value Nonrecursive (List.concat (List.map str_of_type type_decls))])
 
 let intf_generator = Deriving.Generator.V2.make_noarg (fun ~ctxt:_ (_, type_decls) ->
   let sig_of_type type_decl =
-    Ast_helper.with_default_loc type_decl.ptype_loc @@
+    Ast_helper.with_default_loc {type_decl.ptype_loc with loc_ghost = true} @@
       fun () -> sig_of_type type_decl
   in
   List.concat (List.map sig_of_type type_decls))
--- a/src_plugins/eq/ppx_deriving_eq.ml
+++ b/src_plugins/eq/ppx_deriving_eq.ml
@@ -196,14 +196,14 @@ let str_of_type ({ ptype_loc = loc } as type_decl) =
 
 let impl_generator = Deriving.Generator.V2.make_noarg (fun ~ctxt:_ (_, type_decls) ->
   let str_of_type type_decl =
-    Ast_helper.with_default_loc type_decl.ptype_loc @@
+    Ast_helper.with_default_loc {type_decl.ptype_loc with loc_ghost = true} @@
       fun () -> str_of_type type_decl
   in
   [Str.value Recursive (List.concat (List.map str_of_type type_decls))])
 
 let intf_generator = Deriving.Generator.V2.make_noarg (fun ~ctxt:_ (_, type_decls) ->
   let sig_of_type type_decl =
-    Ast_helper.with_default_loc type_decl.ptype_loc @@
+    Ast_helper.with_default_loc {type_decl.ptype_loc with loc_ghost = true} @@
       fun () -> sig_of_type type_decl
   in
   List.concat (List.map sig_of_type type_decls))
--- a/src_plugins/iter/ppx_deriving_iter.ml
+++ b/src_plugins/iter/ppx_deriving_iter.ml
@@ -129,14 +129,14 @@ let sig_of_type type_decl =
 
 let impl_generator = Deriving.Generator.V2.make_noarg (fun ~ctxt:_ (_, type_decls) ->
   let str_of_type type_decl =
-    Ast_helper.with_default_loc type_decl.ptype_loc @@
+    Ast_helper.with_default_loc {type_decl.ptype_loc with loc_ghost = true} @@
       fun () -> str_of_type type_decl
   in
   [Str.value Recursive (List.concat (List.map str_of_type type_decls))])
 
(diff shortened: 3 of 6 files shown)


sim642:
I'm wondering if this is really the best approach. For one, there are other places even in these plugins that access various locations instead of using the default one, so those would still be wrong.
For another, this would mean fixing every other ppx in the same manner. Perhaps it would be better if ppxlib did it once and for all? Or maybe that's not possible for some reason?

sim642:
> For another, this would mean fixing every other ppx in the same manner. Perhaps it would be better if ppxlib did it once and for all? Or maybe that's not possible for some reason?

Apparently ppxlib nowadays does recommend to do it this way: https://ocaml.org/p/ppxlib/0.38.0/doc/good-practices.html#resp_loc. I haven't seen any ppx actually do that.

Tbh, I find the Merlin siblings example there confusing: it's not very clear what this really implies working differently. That description of seeing through the ghost locations is kind of different from the one quoted from the documentation.
Or does that not really matter for Merlin because ppxlib (?) wraps it all into a `[@merlin.hide]` anyway?

fantazio:
> I'm wondering if this is really the best approach. For one, there are other places even in these plugins that access various locations instead of using the default one, so those would still be wrong.
For another, this would mean fixing every other ppx in the same manner.

I agree. Trying to fix ppx-es individually will not scale well. I only found out about this issue with ppx_deriving because it is used in the dead_code_analyzer's testsuite. "Reactive" fixes are enough at the moment in my situation but a more general solution would be preferable.
Maybe ppxlib could offer a `with_default_ghost_loc`  similar to `with_default_loc` but with the extra `{loc with loc_ghost = true}` embedded ?

Note: this issue is not blocking for the dead_code_analyzer, the testsuite can tolerate false positives, and reporting dead generated code is unexpected but lies within the undocumented area.

--------
> Tbh, I find the Merlin siblings example there confusing: it's not very clear what this really implies working differently. That description of seeing through the ghost locations is kind of different from the one quoted from the documentation.

I think that what is meant is that a ghost node is simply treated as if its content was inlined. Thus, although a ghost node may wrap content, thus deepening the AST, the content is considered at the same depth as the ghost, making it sibling with what would have been the ghost's siblings.
My understanding is that this is coherent with [`Location.t`'s documentation](https://ocaml.org/manual/5.5/api/compilerlibref/Location.html#TYPEt) example. If `fun x -> x` is marked as loc_ghost, then the parameter `x` would be considered sibling with `f` (i.e. at the same depth within the same value_binding node), as they are in the actual code. If `fun x -> x` is not marked as loc_ghost, then parameter `x` is deeper than `f`, thus not a sibling.

> Or does that not really matter for Merlin because ppxlib (?) wraps it all into a [@merlin.hide] anyway?

I don't think ppxlib wraps it all into a `[@merlin.hide]` but it does provide helpers. Semantically, there seems to be a difference : a loc_ghost is ignored but traversed while a merlin.hidden node is not traversed, effectively discarding the whole subtree from merlin's analysis.

If someone knowledgeable about merlin could confirm that would be nice.

sim642:
According to [Sherlocode](https://sherlocode.com/?q=loc_ghost), there are ppx-s which already do this kind of thing. Many of them seem to be Jane Street's, so I guess there's some good reasons for using `loc_ghost = true` even though the benefits/reasons aren't very clear.

> Maybe ppxlib could offer a `with_default_ghost_loc` similar to `with_default_loc` but with the extra `{loc with loc_ghost = true}` embedded ?

I doubt that. Ppxlib considers `Ast_helper` (which contains the `default_loc` stuff) to be "old" and recommends `Ast_builder` instead (which has very explicit location specification).
I now realized that there are cases where ppx-s might want to still have non-ghost locations: for example, the argument inside `[@equal …]` would be included in the middle of ppx-generated code but probably wouldn't be ghost because it literally exists in the source file.
I suppose this is very location-specific, so maybe the `Ast_builder` approach which eliminates the `default_loc` would also be good for thinking each instance through.

> Note: this issue is not blocking for the dead_code_analyzer, the testsuite can tolerate false positives, and reporting dead generated code is unexpected but lies within the undocumented area.

It's good to know that this isn't blocking.

Ppxlib does some funny things for usage of generated code though: it creates unit functions with the derived type as argument (which counts as an usage of the type) and dummy definitions which just use the generated functions (which count as usages of those). But I guess that's just enough for the compiler to not complain, a tool like dead_code_analyzer might not want to consider those as proper usages.
It's another thing I haven't fully understood why ppxlib generates those instead of just adding warning suppression attributes (maybe that logic is older than those attributes?). Regardless, I'm not sure if that is a good thing: if the derived function is unused and I want the compiler to warn about it, then it seems reasonable to also know that some deriving clause is pointless.

> [...] the content is considered at the same depth as the ghost, making it sibling with what would have been the ghost's siblings.

But what does such AST mean? It's not even guaranteed to be a valid OCaml AST.
I understand the `Location.t` documentation about it, but not what Merlin's sibling business is about.
Anyway, I do fully see the reason why there should be ghost locations for many/most of the nodes in generated code.

Also, `Location.t` documentation says "ocamlprof skips instrumentation of ghost nodes" which I find to be yet another mismatching interpretation of ghost locations. It can be _very_ relevant to profile ppx-generated code (https://github.com/sim642/ppx_deriving_hash/pull/7) instead of hiding it from profiling results, which would be extremely misleading.
But that's beside the point.

> Semantically, there seems to be a difference : a loc_ghost is ignored but traversed while a merlin.hidden node is not traversed, effectively discarding the whole subtree from merlin's analysis.

Indeed they are two different things, but most of the ppx code I've seen has `[@merlin.hide]` and if that's fully ignored then I guess whether the inner nodes have ghost locations or not wouldn't be making a difference to Merlin anyway. And that might be why this issue has gone unnoticed so long.

sim642 (approved):

Issue #310, opened 2026-07-06
enum plugin sub-error prints extra empty message


sim642:
> [...]
> And the enum plugin actually uses a sub-error, although the output is strange. I'll open some issues about this.
> 
> _Originally posted by @sim642 in https://github.com/ocaml-ppx/ppx_deriving/pull/309#discussion_r3530212034_

```ocaml
File "src_test/enum/test_deriving_enum.ml", line 16, characters 26-28:
16 | type vm = Am [@value 1] | Bm [@value 1] | Cm [@@deriving enum, show]
                               ^^
Error: enum: duplicate value 1 for constructor Bm
File "src_test/enum/test_deriving_enum.ml", line 16, characters 10-12:
16 | type vm = Am [@value 1] | Bm [@value 1] | Cm [@@deriving enum, show]
               ^^

File "src_test/enum/test_deriving_enum.ml", line 16, characters 10-12:
16 | type vm = Am [@value 1] | Bm [@value 1] | Cm [@@deriving enum, show]
               ^^
Error: Same as for Am
```

This is on OCaml 5.5 at least. It might've been fine before 5.3.

gasche:
On first look my impression is that the bug is not in the plugin code, but in another layer that prints the location of suberror twice. Maybe the compiler, maybe dune.

sim642:
When I looked at it, I didn't get to the bottom of it, but my guess would be that it's this compatibility code:
https://github.com/ocaml-ppx/ppx_deriving/blob/3bd1217c90b8cb17de9a4848b7f7d7fd20fac8dd/src/api/ppx_deriving.cppo.ml#L176-L183
I'm not familiar with the `Format_doc` stuff but I have a feeling that something here might be eagerly printing in the middle of some surrounding formatting code.

gasche:
That sounds very plausible, thanks! I'll have a look at tweaking this.

gasche:
I could reproduce the issue with 5.2.0, so it appears that this is not simply the >=5.3 conditional logic that is responsible.

sim642:
Before looking at #315 I tried this on 4.14 as well and there it's even more wack:
```
File "src_test/enum/test_deriving_enum.ml", line 5, characters 26-28:
5 | type va = Aa [@value 1] | Ba [@value 1] | Ca [@@deriving enum, show]
                              ^^
Error: enum: duplicate value 1 for constructor Ba
[...]

PR #314, opened 2026-07-22
enum: better error on value conflicts

gasche:
cc @sim642 

In the context of #310 I looked at the error-producing code in enum, and my impression is that it can improved a bit.

Note: we don't currently have static-error tests in the testsuite, and I was not sure how to add them properly, so for now there are no tests of this behavior.

The new error on my machine is as follows:

```
File "src_test/enum/test_deriving_enum.ml", lines 44-48, characters 0-23:
44 | type conflict =
45 |   | A [@value 1]
46 |   | B [@value 1]
47 |   | C
48 | [@@deriving enum, show]
Error: enum: duplicate value 1 for constructors A, B.
File "src_test/enum/test_deriving_enum.ml", line 45, characters 4-5:
45 |   | A [@value 1]
         ^
  
File "src_test/enum/test_deriving_enum.ml", line 45, characters 4-5:
45 |   | A [@value 1]
         ^
Error: Declaration of A
File "src_test/enum/test_deriving_enum.ml", line 46, characters 4-5:
46 |   | B [@value 1]
         ^
  
File "src_test/enum/test_deriving_enum.ml", line 46, characters 4-5:
46 |   | B [@value 1]
         ^
Error: Declaration of B
```

You can see that the issue of #310 remains, each sub-error has its location printed twice. I think that it would be easy to argue to remove the sub-errors to make the problem go away (especially now that I did the work of collecting the names of the faulty constructors in the main part of the error). But then I would like the OCam compiler ecosystem to grow good support for suberrors (so that we can eventually get the cute Rust ascii art that makes people irrationally happy about error messages), so I am more tempted to try to find which layer of the ecosystem is to blame for this minor bug and then get it fixed.

### DIFF
--- a/src_plugins/enum/ppx_deriving_enum.ml
+++ b/src_plugins/enum/ppx_deriving_enum.ml
@@ -49,24 +49,63 @@ let mappings_of_type type_decl =
           | Rtag (name, true, []) ->
             map acc mappings rtag_attr_value row_field name
           | Rtag _ -> error_arguments loc
-)
+      )
         (0, []) constrs
     | _ -> raise_errorf ~loc:type_decl.ptype_loc
                         "%s can be derived only for variants" deriver
   in
-  let rec check_dup mappings =
-    match mappings with
-    | (a, { txt=atxt; loc=aloc }) :: (b, { txt=btxt; loc=bloc }) :: _ when a = b ->
-      let sigil = match kind with `Regular -> "" | `Polymorphic -> "`" in
-      let sub =
-        [Ocaml_common.Location.errorf
-          ~loc:bloc "Same as for %s%s" sigil btxt] in
-      raise_errorf ~sub ~loc:aloc
-                   "%s: duplicate value %d for constructor %s%s" deriver a sigil atxt
-    | _ :: rest -> check_dup rest
-    | [] -> ()
+  let mappings =
+    (* Put the mappings in source order. *)
+    List.rev mappings in
+  let () =
+    (* Check for duplicate mappings. Consider for example:
+         type t =
+         | A [@value 1]
+         | B [@value 1]
+         | C [@value 1]
+         [@@deriving enum]
+
+       We generate an error that looks as follows (the ">" indicate
+       quoted source with locations), with sub-errors/sub-lcoations:
+
+       > type t = [...]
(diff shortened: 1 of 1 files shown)


gasche:
Thanks to the hints of @sim642 in https://github.com/ocaml-ppx/ppx_deriving/issues/310#issuecomment-5044539495, I could understand the issue with the double-printing of locations in submessages. The issue comes from a mismatch of expectations between our API for submessages (which is not great) and the API of Location in the compiler for submessages (which is not great, but different).

The Location API expects a submessage that does not print its own location (it helpfully takes care of this), but we used the convenience function `Location.print_report` to create the submessage, which does print its own location, and so locations were printed twice.

The fix is slightly different before and after 5.3.0 but it is basically the same idea. The >=5.3.0 fix drops the "footnote" information (which is new in the Location datatype) because I timed out before finding a convenient way to print it, but I thought that it would be an acceptable tradeoff for now (it is _probably_ the case that no one uses it for now anyway, certainly not builtin derivers).

gasche:
New output, before or after 5.3.0:

```
File "src_test/enum/test_deriving_enum.ml", lines 44-48, characters 0-23:
44 | type conflict =
45 |   | A [@value 1]
46 |   | B [@value 1]
47 |   | C [@value 1]
48 | [@@deriving enum, show]
Error: enum: duplicate value 1 for constructors A, B, C.
File "src_test/enum/test_deriving_enum.ml", line 45, characters 4-5:
45 |   | A [@value 1]
         ^
  Declaration of A
File "src_test/enum/test_deriving_enum.ml", line 46, characters 4-5:
46 |   | B [@value 1]
         ^
  Declaration of B
File "src_test/enum/test_deriving_enum.ml", line 47, characters 4-5:
47 |   | C [@value 1]
         ^
  Declaration of C
```

gasche:
I realized on my evening bike ride that the two changes are completely independent, so I will submit a separate PR for the double-location-printing bug (and removed it from this one).

sim642:
I fixed the git merge conflict I caused with #316 and made a few minor tweaks.

gasche:
I think this is okay. (My favorite style would be to call the variable `rev_mappings` as long as they are reverse, to avoid mixing things. We could add a comment to point out that the iteration order from last to first, when building the by-value groups, ensures that each group is in source order.)

PR #315, opened 2026-07-22
fix the double printing of locations in submessages

### LINKED ISSUE #310: enum plugin sub-error prints extra empty message

gasche:
fixes #310

Explanations from https://github.com/ocaml-ppx/ppx_deriving/pull/314#issuecomment-5048652613 where this fix first appeared:

> Thanks to the hints of @sim642 in [#310 (comment)](https://github.com/ocaml-ppx/ppx_deriving/issues/310#issuecomment-5044539495), I could understand the issue with the double-printing of locations in submessages. The issue comes from a mismatch of expectations between our API for submessages (which is not great) and the API of Location in the compiler for submessages (which is not great, but different).
> 
> The Location API expects a submessage that does not print its own location (it helpfully takes care of this), but we used the convenience function `Location.print_report` to create the submessage, which does print its own location, and so locations were printed twice.
> 
> The fix is slightly different before and after 5.3.0 but it is basically the same idea. The >=5.3.0 fix drops the "footnote" information (which is new in the Location datatype) because I timed out before finding a convenient way to print it, but I thought that it would be an acceptable tradeoff for now (it is _probably_ the case that no one uses it for now anyway, certainly not builtin derivers).

### DIFF
--- a/src/api/ppx_deriving.cppo.ml
+++ b/src/api/ppx_deriving.cppo.ml
@@ -171,21 +171,26 @@ let lookup name =
 let raise_errorf ?sub ?loc fmt =
   let module Location = Ocaml_common.Location in
   let raise_msg str =
-    let sub =
-      let msg_of_error err =
 #if OCAML_VERSION >= (5, 3, 0)
-        let loc = err.Location.main.loc in
-        let print_report fmt x =
-          Ocaml_common.Format_doc.deprecated_printer
-            (fun fmt -> Location.print_report fmt x) fmt
-        in
-        Location.msg ~loc "%a" print_report err
+    let collect_subs err acc =
+      err.Location.main ::
+      err.Location.sub @
+      match err.Location.footnote with
+      | None -> acc
+      | Some doc -> Location.mknoloc doc :: acc
+    in
+    let sub = Option.map (fun sub -> List.fold_right collect_subs sub []) sub in
 #else
-        { txt = (fun fmt -> Location.print_report fmt err);
-          loc = err.Location.main.loc }
-#endif
+    let msg_of_error err =
+      let msg fmt =
+        let printer = !Location.report_printer () in
+        printer.pp_main_txt printer err fmt err.Location.main.txt;
+        printer.pp_submsgs printer err fmt err.Location.sub;
       in
-      Option.map (List.map msg_of_error) sub in
+      { txt = msg; loc = err.Location.main.loc }
+    in
+    let sub = Option.map (List.map msg_of_error) sub in
+#endif
     let err = Location.error ?sub ?loc str in
     raise (Location.Error err) in
   Printf.ksprintf raise_msg fmt


gasche:
I just pushed a slightly better >=5.3 version which does not drop the footnotes anymore, and is conceptually simpler (no `Format_doc.deprecated` stuff anymore).

sim642:
The diffs of the errors due to this PR for my example from #310 are the following:

On OCaml 4.14:
```diff                        
 File "src_test/enum/test_deriving_enum.ml", line 5, characters 26-28:
 5 | type va = Aa [@value 1] | Ba [@value 1] | Ca [@@deriving enum, show]
                               ^^
@@ -6,7 +6,4 @@
 File "src_test/enum/test_deriving_enum.ml", line 5, characters 10-12:
 5 | type va = Aa [@value 1] | Ba [@value 1] | Ca [@@deriving enum, show]
               ^^
-  File "src_test/enum/test_deriving_enum.ml", line 5, characters 10-12:
-  5 | type va = Aa [@value 1] | Ba [@value 1] | Ca [@@deriving enum, show]
-                ^^
-  Error: Same as for Aa
+  Same as for Aa
```

On OCaml 5.5:
```diff                 
 File "src_test/enum/test_deriving_enum.ml", line 5, characters 26-28:
 5 | type va = Aa [@value 1] | Ba [@value 1] | Ca [@@deriving enum, show]
                               ^^
@@ -7,8 +6,4 @@
 File "src_test/enum/test_deriving_enum.ml", line 5, characters 10-12:
 5 | type va = Aa [@value 1] | Ba [@value 1] | Ca [@@deriving enum, show]
               ^^
-  
-File "src_test/enum/test_deriving_enum.ml", line 5, characters 10-12:
-5 | type va = Aa [@value 1] | Ba [@value 1] | Ca [@@deriving enum, show]
-              ^^
-Error: Same as for Aa
+  Same as for Aa
```

So besides the double location (#310 itself) and the indentation (https://github.com/ocaml-ppx/ppx_deriving/issues/310#issuecomment-5061347398), it also fixes a difference between the OCaml versions regarding the extra empty line.
And finally it removes the `Error: ` prefix on the submessage which also matches the submessages used by the compiler itself.
So now the outputs actually match between OCaml versions.

(While looking at the `Location` code in the trunk compiler, I found out that `loc_ghost` actually does something: it hides the location of a submessage. That's quite useful to know since I was planning to extend #313 to also the non-default locations.)

sim642 (approved):

PR #316, opened 2026-07-24
Mark more locations as `loc_ghost`

sim642:
This is a continuation of #313.
There are plenty of other places in the plugins which take a location from some subnode, rather than `default_loc`, and new nodes created with that location should also be `loc_ghost`.

In order to not go too crazy, I omitted this for the locations that are only used for `raise_errorf` because the semantics of `loc_ghost` on those exceptions isn't clear to me: they don't end up as locations of new `Parsetree` nodes.
The situation might be a bit more hazy for `extension_errorf` (which ppx_deriving doesn't use much): they do end up as `Parsetree` nodes, but they would always mean something ghost-like (except if the program literally contains `[@ocaml.error ...]` but that would be very strange).

### DIFF
--- a/src_plugins/create/ppx_deriving_create.ml
+++ b/src_plugins/create/ppx_deriving_create.ml
@@ -35,6 +35,7 @@ let find_main labels =
     (None, []) labels
 
 let str_of_type ({ ptype_loc = loc } as type_decl) =
+  let loc = {loc with loc_ghost = true} in
   let quoter = Ppx_deriving.create_quoter () in
   let creator =
     match type_decl.ptype_kind with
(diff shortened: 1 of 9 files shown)


sim642:
@fantazio I suppose most locations were already covered by #313 but perhaps this makes a little bit of a further difference to dead_code_analyzer in some cases.

sim642:
I'll just merge this small change as well so I can do a new release with the many recent fixes.

fantazio:
Sorry for the delay, I was on holidays.
Thanks for taking care of it.

> @fantazio I suppose most locations were already covered by https://github.com/ocaml-ppx/ppx_deriving/pull/313 but perhaps this makes a little bit of a further difference to dead_code_analyzer in some cases.

Probably yes. The tests are currently limited to the use of `ppx_deriving.eq`, with the intent to check that uses of user code within generated code are properly tracked. Checking that generated code is not tracked is a (beneficial) side effect.
It might be worth extending the dead_code_analyzer testsuite to more ppx_deriving constructs, at least to keep track of potential "false positives" like [the ones discovered with 6.1.2](https://github.com/LexiFi/dead_code_analyzer/pull/87).

